PR #13, opened 2015-07-28
use raw strings for generated code

jakemac53:
closes https://github.com/dart-lang/reflectable/issues/12


jakemac53:
Also fixed up the test which had some type errors, is this being ran on the package bots?

eernstg:
Very good, we'll make it a rule to use raw strings for generated code!

About running in checked mode: We use a script, `bottest`, which is intended to emulate a large subset of the procedure on the package bots as precisely as possible (in particular, it runs the same 'annotated_steps.py'), so I'm surprised that it isn't running checked mode already (during `bottest` and on the package bots). For now, I've changed the Makefile goals used to run smaller/quicker test procedures to run in checked mode in my local repo, such that it will be effective from the next `git cl land`.

Merging now.

PR #14, opened 2015-07-28
Broken link when opened from https://github.com/dart-lang/reflectable/blob/master/reflectable/doc/TheDesignofReflectableCapabilities.pdf

zoechi:


eernstg:
Thanks!

PR #18, opened 2015-08-06
correct _LibraryMirrorImpl.invoke

luisvt:
use `new Symbol(memberName)` instea…d of just `memberName`

### DIFF
--- a/reflectable/lib/src/reflectable_implementation.dart
+++ b/reflectable/lib/src/reflectable_implementation.dart
@@ -227,7 +227,7 @@ class _LibraryMirrorImpl extends _DeclarationMirrorImpl
   Object invoke(String memberName, List positionalArguments,
       [Map<Symbol, dynamic> namedArguments]) {
     if (!reflectableSupportsStaticInvoke(_reflectable, memberName,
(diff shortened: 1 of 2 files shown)


luisvt:
I signed it!

eernstg:
Hello Luis,

thanks for your input! It is correct that the plain `String` should be turned into a `Symbol`. However, the reason why this had not surfaced as a problem with the current test suite is that support for top level entities (including library mirrors) have not yet been implemented in transformed code. This means that it is useful to separate the test of top level invocation out from instance level invocation, such that we can run the test for the latter in both untransformed and transformed code, and mark the former as "expected to fail" for transformed code until we implement it (which is near the top of the priority list, anyway). I've fetched your pull request to my desktop, made that adjustment, and then I'll push the result as soon as all the pieces fit together again.

sigurdm:
The modified version of this CL is at: https://codereview.chromium.org/1281543002

eernstg:
We discussed this a bit, and agreed that it's a better approach to accept the pull request as is, and then make the adjustments, such that the pull request gets the right attribution in the first place. So we'll do that. ;-)

PR #22, opened 2015-09-02
add test for getting return type of a getter

jakemac53:
This currently fails, but demonstrates the use case that polymer needs.

### DIFF
--- a/test_reflectable/test/implicit_getter_setter_test.dart
+++ b/test_reflectable/test/implicit_getter_setter_test.dart
@@ -24,6 +24,7 @@ class A {
   static A f3;
   static final A f4 = new A(null);
   static const f5 = 542;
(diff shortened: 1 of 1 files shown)


eernstg:
Getting the return type of a getter should be corrected by https://codereview.chromium.org/1311873004/.

jakemac53:
Should this test now be merged then? Either that or this issue can be closed if its tested elsewhere.

jakemac53:
whoops didn't mean to close :)

eernstg:
I added material similar to your input to 'parameter_test.dart' (where other tests of explicitly declared getters and setters occur), and I also added tests on the return types of the already declared implicit getters and setters in 'implicit_getter_setter_test.dart', so it should be covered today. Thanks again for your input, it was indeed missing at that point!

PR #46, opened 2015-11-02
Add isEnum to reflectable ClassMirror

luisvt:
Add the `isEnum` method to reflectable `ClassMirror` class. In this way we will be able to check if the object is an enum

### DIFF
--- a/reflectable/lib/mirrors.dart
+++ b/reflectable/lib/mirrors.dart
@@ -333,6 +333,12 @@ abstract class ClassMirror implements TypeMirror, ObjectMirror {
   ClassMirror get superclass;
   List<ClassMirror> get superinterfaces;
   bool get isAbstract;
+
+  /**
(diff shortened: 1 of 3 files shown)


sigurdm:
Thanks for the contribution!

Would you mind rebasing this to be a single commit? Then we can merge it and afterwards put in an implementation in the transformer.

luisvt:
ok

luisvt:
done, there is only one commit now

eernstg:
Thanks a lot! I'm currently debugging other stuff, but I'll look into this as soon as possible.

eernstg:
With the CLA issue resolved, we can finally handle this PR! ;-)

The PR was merged manually, because of conflicts that had arisen in the mean time. The resulting commit is 63109c6b6372e5c5ee293b46fcd0054256633616.

PR #99, opened 2017-10-30
Use package build rather than a pub transformer

eernstg:


reflectable/bin/reflectable_builder.dart
```diff
@@ -0,0 +1,39 @@
...
+/// points to transform.
+
+library reflectable.bin.reflectable_builder;
+
+import 'package:barback/src/transformer/barback_settings.dart';
+import 'package:build_barback/build_barback.dart';
+import 'package:build_runner/build_runner.dart';
+import 'package:reflectable/transformer.dart';
+
+main(List<String> arguments) async {
```
jakemac53: Note that you could get the root package this way:

```dart
var graph = new PackageGraph.forThisPackage();
var package = graph.root.name;
```

reflectable/README.md
```diff
@@ -123,98 +113,52 @@ main() {
...
-completely under the programmer's control.
-
-The stand-alone version of the reflectable transformer is
-`bin/reflectable_transformer.dart`. It can be invoked using
-`pub global run` (after activating it with
-`pub global activate reflectable`):
+You may also wish to specify constraints on the version, depending on the
+approach to version management that your software otherwise employs.
 
+Please make sure that the reflectable package has been made active:
```
jakemac53: Note that while it is totally valid to provide this pre-wrapped executable, we also want to support people running multiple `BuildAction`s with the same build script, so its probably worth exposing the reflectable builder directly and showing an example of how to use it (you could just point at `bin/reflectable_builder.dart` even).

jakemac53:
I left a few notes, but overall this looks good! Thanks for taking the time to convert this over.

As a general note, the way these builders are actually set up and ran is still in flux a bit, (we would like to autogenerate these build scripts based on dependencies, plus optional configuration). We will let you know as things change though, and could add an e2e test for reflectable into the build repo if you are interested in that.

Another note is that the build package does do incremental builds, but as a single global step this won't be able to take much advantage of that. It might be worth exploring more incremental approaches in the future, depending on how long this action takes to run for people in practice. It might not be feasible anyways though given the somewhat global nature of some of the features in this package.

reflectable/README.md
```diff
@@ -123,98 +113,52 @@ main() {
...
+```shell
+...> pub global activate reflectable
```
jakemac53: You might not want to suggest `global activate`, and instead just regular `pub run`. You probably want to make sure the version of the reflectable builder is the same as the version of reflectable the package is depending on.

test_reflectable/test/exported_main_lib.dart
```diff
@@ -4,9 +4,9 @@
```
sigurdm: Perhaps document the choice between:
- In the importing library, make an actual main calling `initializeReflectable` and the imported main
- modifying the library exporting the main.

reflectable/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
 
+## 2.0.0
+
+* Switched to a new technology stack: Reflectable no longer provides a pub
+  transformer, it uses package `build` to generate code as a separate
+  step. This is a **breaking change** for *every* program using reflectable
+  because it requires a different workflow, it requires the generated code to
+  be imported explicitly by the root library (the one that contains the `main`
+  function), and it requires invocation of `initializeReflectable()` at the
+  beginning of `main`.
```
sigurdm: Perhaps also mention that this requires changes to the build setup, and a reference to a description of the commands to run to generate the code.

test_reflectable/test/exported_main_lib.dart
eernstg: I added a comment which should describe these things.

reflectable/bin/reflectable_transformer.dart
```diff
@@ -50,7 +50,7 @@ transform(String package, String entryPoint, Map<String, String> inputs) async {
 
     TestDeclaringTransform declaringTransform =
         new TestDeclaringTransform(inputName);
-    await transformer.declareOutputs(declaringTransform);
+    transformer.declareOutputs(declaringTransform);
```
eernstg: I did not intend to remove that `await`, it's been re-added.

reflectable/bin/reflectable_builder.dart
```diff
@@ -0,0 +1,39 @@
...
+        "argument; exiting.");
+  } else {
+    String packageName = arguments[0];
+    List<String> filePaths = arguments.sublist(1);
+    // TODO(eernst): Sort out which settings to use here.
+    BarbackSettings settings = new BarbackSettings({
+      "entry_points": [filePaths],
+      "formatted": true,
+    }, BarbackMode.DEBUG);
+    final reflectableTransformerBuilder = new TransformerBuilder(
```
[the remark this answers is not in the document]
eernstg: I was thinking about that too, but let's take that step separately.

reflectable/bin/reflectable_builder.dart
eernstg: Great, I've been looking for that! Done.

reflectable/README.md
eernstg: Done.

eernstg:
> the way these builders are actually set up and ran is still in flux a bit .. We will let you know

OK, it would be very nice to get a heads up when this area develops! 

> could add an e2e test for reflectable into the build repo if you are interested in that

An end-to-end test? Sounds like a good idea, unless it's redundant because the buildbot for package test_reflectable already does the same thing.

PR #107, opened 2018-01-12
update to latest package:build and add custom build.yaml

jakemac53:
Here is an example of moving reflectable to the new package:build_runner config.

After this cl, most users of package:reflectable need only do `pub run build_runner build`, and reflectable will automatically bootstrap their application. They would still need to know to include the reflectable.dart file instead of the original one though.

 We have some rough docs [here](https://github.com/dart-lang/build/tree/master/build_config#defining-builders-to-apply-to-dependents-similar-to-transformers) describing what this all does and are working on better ones.

See the example build.yaml I added to the test_reflectable package as well. Users can use the `generate_for` argument to change the primary inputs for the builder using that (this would be similar to the arguments the current build script accepts).

Let me know if you have any questions!

### DIFF
--- a/reflectable/lib/reflectable_builder.dart
+++ b/reflectable/lib/reflectable_builder.dart
@@ -6,9 +6,22 @@ library reflectable.src.reflectable_builder;
 
 import 'dart:async';
 import 'package:barback/src/transformer/barback_settings.dart';
+import 'package:build/build.dart';
+import 'package:build_config/build_config.dart';
 import 'package:build_barback/build_barback.dart';
 import 'package:build_runner/build_runner.dart';
 import 'package:reflectable/transformer.dart';
+
+TransformerBuilder reflectableBuilder(BuilderOptions options) {
+  var config = new Map<String, Object>.from(options.config);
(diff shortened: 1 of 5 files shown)


reflectable/lib/reflectable_builder.dart
```diff
@@ -6,19 +6,30 @@ library reflectable.src.reflectable_builder;
...
 import 'dart:async';
 import 'package:barback/src/transformer/barback_settings.dart';
+import 'package:build/build.dart';
+import 'package:build_config/build_config.dart';
 import 'package:build_barback/build_barback.dart';
 import 'package:build_runner/build_runner.dart';
 import 'package:reflectable/transformer.dart';
 
+TransformerBuilder reflectableBuilder(BuilderOptions options) {
+  var config = new Map<String, Object>.from(options.config);
```
jakemac53: Just as an fyi, this corresponds to the `options` field in a custom build.yaml file, if a package has one. So for instance they could set up:

```yaml
targets:
  my_package:
    builders:
      reflectable:
        options:
          formatted: true
```

PR #109, opened 2018-02-28
Flatten reflectable repo to only contain the reflectable package

eernstg:
This commit changes the reflectable repository from being a set of three packages (and a few extra files like the top-level `README.md`) to a state where `test_reflectable` and `coverage_reflectable` have been deleted, and `reflectable`-the-package is the same thing as `reflectable`-the-repository.

The point is that this allows the new `test_reflectable` repository to refer to `reflectable` as a regular dependency, such that we can make the package buildbots run the tests.


lib/capability.dart
```diff
@@ -475,7 +475,7 @@ class _AdmitSubtypeCapability implements ReflecteeQuantifyCapability {
 
 /// Thrown when reflection is invoked without sufficient capabilities.
 abstract class NoSuchCapabilityError extends Error {
-  factory NoSuchCapabilityError(message) = _NoSuchCapabilityErrorImpl;
+  factory NoSuchCapabilityError(String message) = _NoSuchCapabilityErrorImpl;
```
sigurdm: Perhaps leave this change for a separate CL

lib/src/incompleteness.dart
```diff
@@ -21,7 +21,7 @@ class UnreachableError extends Error {
 /// `throw unreachableError(..)`, which will indicate to the compiler that this
 /// expression throws (even though it actually happens here). This way we avoid
 /// warnings about a missing return problem, and the code may be more readable.
-void unreachableError(String message) {
+Null unreachableError(String message) {
```
sigurdm: Perhaps leave this change for a separate CL

lib/src/incompleteness.dart
```diff
@@ -33,7 +33,7 @@ void unreachableError(String message) {
 /// this expression throws (even though it actually happens here). This way we
 /// avoid warnings about a missing return problem, and the code may be more
 /// readable.
-void unimplementedError(String message) {
+Null unimplementedError(String message) {
```
sigurdm: Perhaps leave this change for a separate CL

lib/src/transformer_implementation.dart
```diff
@@ -238,11 +238,25 @@ class ClassElementEnhancedSet implements Set<ClassElement> {
   Iterable<T> map<T>(T f(ClassElement element)) =>
       classElements.items.map<T>(f);
 
+  @override
```
sigurdm: Same in this file

PR #112, opened 2018-02-28
Adjusted .status to match the new directory structure

eernstg:
TBR: During the 'flattening' of reflectable, the test files were moved from `test/mock_tests` to `test`. This CL updates `.status` to match that change.

### DIFF
--- a/.status
+++ b/.status
@@ -6,10 +6,10 @@ test/*: SkipByDesign # Testing in build to get a package root.
 
 # Prevent execution of non-test files with name *_test.dart in packages.
 build/test/packages: SkipByDesign
-build/test/mock_tests/packages: SkipByDesign
+build/test/packages: SkipByDesign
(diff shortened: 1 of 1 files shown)


.status
```diff
@@ -6,10 +6,10 @@ test/*: SkipByDesign # Testing in build to get a package root.
 
 # Prevent execution of non-test files with name *_test.dart in packages.
 build/test/packages: SkipByDesign
-build/test/mock_tests/packages: SkipByDesign
+build/test/packages: SkipByDesign
```
sigurdm: Is this not the same line twice?

eernstg: Indeed! They used to be different (and the `build/test/packages` line used to have no effect because there were no tests in `test`, only in `test/mock_tests`); I'm removing the redundant line in the next CL.

PR #114, opened 2018-03-06
Pub run test mar18

eernstg:
Enabled `pub run test` (which runs locally), but I cannot test a couple of things, e.g., that the given '.test_config' file will actually make the buildbot run the tests with that command. Nevertheless, I'd like to try it out, and it's red now anyway. ;-)


lib/test_transform.dart
```diff
@@ -10,8 +10,12 @@ import 'dart:convert';
 import 'dart:io' as io;
 
 import 'package:barback/barback.dart';
+import 'package:package_resolver/package_resolver.dart';
+import "package:package_config/packages_file.dart" as packages_file;
 import 'package:source_span/source_span.dart';
 
+Uri dotPackages = io.Platform.script;
```
sigurdm: Is this used?

lib/test_transform.dart
```diff
@@ -33,43 +36,37 @@ class TestTransform implements Transform {
...
     _primaryAssetId = new AssetId.parse(fileName);
     _primaryAsset = new Asset.fromString(_primaryAssetId, fileContents);
     assets[_primaryAssetId] = _primaryAsset;
+    Uri scriptUri = io.Platform.script;
+    String uriDataScriptString = scriptUri.data?.contentAsString();
+    if (uriDataScriptString != null) {
+      // `uriDataScriptString` actually contains the entire program, which
+      // imports the test file using a `file:///` uri.
+      int startIndex = uriDataScriptString.indexOf("file:///");
+      if (startIndex == 0) {
```
sigurdm: should be -1

lib/test_transform.dart
```diff
@@ -33,43 +36,37 @@ class TestTransform implements Transform {
...
+      this.fileName, this.fileContents, this.package) {
     _primaryAssetId = new AssetId.parse(fileName);
     _primaryAsset = new Asset.fromString(_primaryAssetId, fileContents);
     assets[_primaryAssetId] = _primaryAsset;
+    Uri scriptUri = io.Platform.script;
+    String uriDataScriptString = scriptUri.data?.contentAsString();
+    if (uriDataScriptString != null) {
+      // `uriDataScriptString` actually contains the entire program, which
+      // imports the test file using a `file:///` uri.
+      int startIndex = uriDataScriptString.indexOf("file:///");
```
sigurdm: Perhaps use a regexp including the whole line you want to match.

lib/test_transform.dart
```diff
@@ -91,15 +99,14 @@ class TestTransform implements Transform {
...
-      path = [packageRoot, id.package, pathWithoutLib]
-          .join(io.Platform.pathSeparator);
+      path = (await packageResolver.urlFor(id.package, pathWithoutLib))
+          .toFilePath();
     }
     io.File file = new io.File(path);
-    if (!(await file.exists())) {
-      throw new AssetNotFoundException(id);
-    }
+    if (!(await file.exists())) throw new AssetNotFoundException(id);
```
sigurdm: Is this an intended change?

lib/test_transform.dart
```diff
@@ -91,15 +99,14 @@ class TestTransform implements Transform {
     if (package == id.package) {
       path = id.path;
     } else {
+      if (packageResolver == null) await _setupPackageResolver();
```
sigurdm: Perhaps move the caching logic into an async getter?
```dart
Future<PackageResolver> get packageResolver() async {
  if (_packageResolver == null) {
    await _setupPackageResolver();
  }
  return _packageResolver;
}
```

lib/test_transform.dart
eernstg: Yes, I just didn't see the need to keep the block.

lib/test_transform.dart
eernstg: Ah, that's clearly better (and it gets to do what _setupPackageResolver used to do, there is no point in having that separately). Using a method, because getters are often assumed to "finish quickly".

lib/test_transform.dart
eernstg: Right, and that would only match a single line, so we wouldn't get the wrong result even if we were to receive this:
```dart
...
import 'file:///whatever/I-am-just-cheating-you.dart';
import 'file:///this/is/the/one/we/want/test_reflectable/test/some_test.dart';
...
```
Done.

lib/test_transform.dart
eernstg: Nope, forgot to delete it when the inst-var with the same name was added. Thanks!

Issue #136, opened 2018-07-02
Please update dependencies


MikeMitterer:
Hi, could you please update your build_runner dependencies?
8.3 always crashes on my Mac.
Thanks

eernstg:
Simple updates do not suffice. Looking into it.

MikeMitterer:
I cloned your lib changed those deps:

```
  build_config: ^0.3.0
  build_runner: ^0.9.0
  dart_style: 1.1.1
```
At least all the test in dryice with upgraded dependencies are on green...

MikeMitterer:
Booooom - now I know what you mean with "bug"... `pub run build_runner build` crashed. My dryice-tests must have used a cached version or so

eernstg:
Running with SDK (and tools) version 2.0.0-dev.60.0 works here, and so do 61 &ndash; 63, but from 64 and up it fails. Would you be able to use one of those older versions for now?

eernstg:
I don't get any error messages referring to files in `reflectable`, so I don't see exactly the same things that you see. I'll go home now, and take another look tomorrow.

MikeMitterer:
I switched to 2.xxx.63 - it works. Thanks so far

MikeMitterer:
I thought it works... I had to change the dependencies in you cloned packages to `build_runner: ^0.8.10` otherwise I ran into https://github.com/dart-lang/pub/issues/1932 with
```bash
NoSuchMethodError: Class '_SingleBuild' has no instance method 'runPostProcessBuilder'.
Receiver: Instance of '_SingleBuild'
Tried calling: runPostProcessBuilder(Instance of 'ModuleCleanup', Instance of 'AssetId', Instance of 'SingleStepReader', Instance of 'AssetWriterSpy', Instance of 'BuildForInputLogger', Instance of 'AssetGraph', Instance of 'PostProcessAnchorNode', 467)
``` 
Oh - and before I had to `pub global activate build_runner` to switch build_runner back to 0.8.10
x hours of try and error which Dart version works with wich constraints

eernstg:
I'm currently working on an update of reflectable to eliminate barback (which is deprecated at this point). That should take care of the version conflicts, but many things need to be changed in order to do that so it may not be in a working state right now (and I'm OOO during much of the rest of July&mdash;conference then vacation).

MikeMitterer:
[...]

Issue #130, opened 2018-06-04
Migrate off of code_transformers


natebosch:
We should stop using `barback`, `build_barback`, and `code_transformers` dependencies and migrate to fully `package:build` based workflows including `buildStep.resolver`.

eernstg:
Indeed we should do that! @natebosch, do you know about any documents that would be helpful in order to get started?

natebosch:
I'm not sure how much needs to be migrated. https://github.com/dart-lang/build/blob/master/docs/from_barback_transformer.md is the general document for moving from the `Transformer` interface to the `Builder` interface but it doesn't have a lot of detail. For the most part the APIs are relatively similar and it's the restrictions that end up being the pain points.

eernstg:
@natebosch, thanks a lot!

eernstg:
Doesn't quite work yet, but here is a preview: https://github.com/dart-lang/reflectable/tree/eliminate_barback_jul18.

The generated code may contain ill-formed import statements of the form `import 'asset:...';`, and it seems to work if they are adjusted to `import 'package:..';` in a way which is typically obvious. Other than that, most things seem to work.

eernstg:
At this point the [eliminate_barback_jul18 branch](https://github.com/dart-lang/reflectable/tree/eliminate_barback_jul18) is running all tests successfully, and a review has been requested. Hopefully we can land it soon and make it release 2.0.3.

One issue remains (see [here](https://github.com/dart-lang/sdk/issues/34148)), which may cause some metadata to be ignored (in particular, [this one](https://github.com/dart-lang/test_reflectable/blob/5e9a29c8ee82d96f2be04f228c9b4099c10f50d8/test/meta_reflectors_meta.dart#L14) has no effect). A promising workaround is available (but the exact cause of the problem is not yet known, so there's some guesswork involved): Put another copy of that metadata on an import of `package:reflectable/reflectable.dart` in some other library that imports this library (such that we don't rely on using `AllReflectorsMetaReflector` or a similar class on an import before the analyzer has seen the declaration of that class itself). In the example, it could go [here](https://github.com/dart-lang/test_reflectable/blob/5e9a29c8ee82d96f2be04f228c9b4099c10f50d8/test/meta_reflectors_test.dart#L16).

eernstg:
[...]

PR #142, opened 2018-08-15
Eliminate barback jul18

eernstg:
Please ignore the reflectable/test "mock" tests --- I haven't yet figured out how to update them to a builder context (as opposed to the transformer context that we used to have). Nobody else depends on them, and the package bot is not running any more, so that's not technically a problem right now.

Otherwise, this CL ports reflectable to Dart 2 (#130), and in particular it eliminates all traces of package `barback` and the associated transformers (which is not available any more). Part of this is a change of dependencies (using analyzer 0.32.4, currently, and build_runner 0.10.1, and so on), and that's the reason for this effort in the first place (some users of reflectable couldn't satisfy the dependencies of something bigger that included reflectable: #136).


lib/src/builder_implementation.dart
```diff
@@ -3902,125 +3877,97 @@ ${imports.join('\n')}
...
-    if (currentEntryPoint == null) return;
+    print(_librariesByName); // DEBUG
+    LibraryElement reflectableLibrary =
+        _librariesByName["reflectable.reflectable"];
+
+    if (reflectableLibrary == null) {
+      // Stop and let the original source pass through without changes.
+      log.info("Ignoring entry point $inputId that does not "
+          "include the library 'package:reflectable/reflectable.dart'");
+      if (const bool.fromEnvironment("reflectable.pause.at.exit")) {
```
natebosch: cc @jakemac53 

Note that we don't know what environment variables you read during the build and won't invalidate the output if a new build is run with a different environment. Is it possible to put this argument in the `BuilderOptions` and configure with `build.yaml` instead? If it's changed we'll know to rerun the appropriate build steps.

eernstg: Those things are only used for controlling debug-ish output, so it shouldn't matter for the decision about whether or not to regenerate code.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## 2.0.2
 
+* Eliminate the use of barback (which is now deprecated). This change
+  requires many changes in the implementation, but it should not give
+  rise to incompatibilities for clients of reflectable.
+
+* Update many version dependencies: TODO
```
sigurdm: This should be filled in :)

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+
+## 2.0.2
```
sigurdm: It seems this version now appears twice in the changelog.

lib/reflectable_builder.dart
```diff
@@ -2,25 +2,51 @@
...
+
+    var outputId = inputId.changeExtension('.reflectable.dart');
+
+    // Get the `LibraryElement` for the primary input.
+    var inputLibrary = await buildStep.inputLibrary;
+
+    // Resolve all libraries reachable from the primary input.
+    List<LibraryElement> visibleLibraries = [];
+    await for (LibraryElement library in resolver.libraries) {
+      visibleLibraries.add(library);
```
sigurdm: I think this loop can be simplified to:
```
visibleLibraries = await resolver.libraries.toList();
```

lib/reflectable_builder.dart
```diff
@@ -32,16 +58,11 @@ Future<BuildResult> reflectableBuild(List<String> arguments) async {
...
-      new ReflectableTransformer.asPlugin(settings),
-      const <String, List<String>>{
-        '.dart': const ['.reflectable.dart']
-      },
-    );
+    BuilderOptions options = new BuilderOptions(
+        <String, dynamic>{"entry_points": arguments, "formatted": true},
+        isRoot: true);
+    final builder = new ReflectableBuilder(options);
+    // Plus: new ReflectableTransformer.asPlugin(settings),
```
sigurdm: What does this comment mean?

lib/src/reflectable_builder_based.dart
```diff
@@ -172,14 +170,12 @@ class ReflectorData {
...
-        "reflectable transformer's entry_points in pubspec.yaml?");
+        "reflectable builder's entry_points in pubspec.yaml?");
 
 /// This mapping translates symbols to strings for the covered members.
 /// It will be initialized in the generated code.
 Map<Symbol, String> memberSymbolMap =
-    throw new StateError("Reflectable has not been initialized. "
-        "Did you forget to add the main file to the "
-        "reflectable transformer's entry_points in pubspec.yaml?");
+    throw new StateError("Reflectable has not been initialized.");
```
sigurdm: This message seems less informative now. Could we add some kind of advice for how to initialize reflectable?

CHANGELOG.md
eernstg: Done.

CHANGELOG.md
eernstg: Corrected to 2.0.3, which is a suitable version number for pub.dartlang.org as well.

lib/reflectable_builder.dart
eernstg: Indeed, thanks!

lib/reflectable_builder.dart
eernstg: That comment was just left over from old code. Deleted.

lib/src/reflectable_builder_based.dart
eernstg: Adjusted both messages to be identical (using a single `const String`), giving the advice that `main` should call `initializeReflectable()` as its first action, which would presumably be the typical (and only?) way any developer could end up encountering the situation where those `StateError`s are thrown.

If no code is generated we will have a compile-time error (because there is an import for something that doesn't exist), so the old reason for having this does not exist any more.

If there is no import of the generated file (whether or not it's present) and no `initializeReflectable()` then we may still have other libraries in the same program that import reflectable-something, and in that case they will actually still hit the above `StateError`, so I think we're OK in all cases! ;-)

PR #146, opened 2018-08-22
Imported library null fix aug18

eernstg:
Just a null error bug fix, plus preparation for publishing it as 2.0.4.


eernstg:
I'll land this because it's so tiny, and it would be nice to fix the bug and have a fast response to the submitter of https://github.com/dart-lang/reflectable/issues/145.

PR #148, opened 2018-08-28
Support static type arguments aug18

eernstg:
This is a revival of a branch from 2017, porting it such that it works as part of reflectable 2.0.4. It adds support for some statically resolved type arguments. It is possible to detect via reflection that a given parameter has type `List<int>` or a similar fully resolved type. There is no support for the case where a type annotation contains a formal type parameter from an enclosing function or class (say, the parameter type of `List.addAll` which is declared as `Iterable<E>` where `E` is only known at run time). Attempts to retrieve such a type will cause an `UnimplementedError` to be thrown.

PS: I'll make test_reflectable a properly reviewed repo _real soon now_. However, the test case for this feature is available there now at https://github.com/dart-lang/test_reflectable/blob/master/test/static_type_arguments_test.dart.


lib/src/builder_implementation.dart
```diff
@@ -941,6 +941,8 @@ class _ReflectorDomain {
       classesToAdd.forEach(addClass);
     }
 
+    // Add
```
sigurdm: This seems spurious

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+## 2.0.5
+
+* Enhancement of the support for reflection on type arguments, covering
+  cases where a type annotation is a parameterized type where all type
+  arguments are resolved statically (e.g., `List<int> f();`). We (still) do
+  not support cases where one or more type arguments are or contain
+  type variables from an enclosing function or class (like the type of
+  the formal parameter to `List.addAll`: `List<E>`), and with such type
+  arguments a dynamic error is raised if an attempt is made to access it.
+  Similarly, we (still) do not support extracting the value of the actual
```
sigurdm: does the `(still)` mean that we are going to support it at some point?

README.md
```diff
@@ -278,13 +278,24 @@ the following parts are still incomplete:
...
-  returning the actual type arguments would require runtime support that does
-  not currently exist.
+- Type arguments of generic types are only supported in some statically
+  resolved cases, because we do not have the primitives required for a
+  general implementation. E.g., when it is known statically that a given
+  list of actual type arguments is empty then the empty list is
+  returned. Moreover, if support for reflected types is included and a type
+  annotation contains only types which are fully resolved statically, it is
+  possible to get these type arguments. So, for instance, with a field of
+  type `Map<int, String>` and suitable capabilities, we can invoke
```
sigurdm: I would write the examples in a long form to show full examples of what is supported and what is not. I think that would ease the comprehension. Maybe along the lines of:
````
```dart
class Reflector extends Reflectable {
  const Reflector() : super(declarationsCapability, typeCapability, typeRelationsCapability, reflectedTypeCapability);
}

@Reflector()
class A<T> {
  Map<int, String> map;
  List<X> list;
}

main() {
  initializeReflectable();
  final declarations = reflector.reflectType(A).declarations;
  declarations['map'].reflectedTypeArguments[0]; // Will return int.
  declarations['list'].reflectedTypeArguments[0]; // Will throw at runtime.
}
```
````

lib/src/builder_implementation.dart
eernstg: Deleted, thanks!

CHANGELOG.md
eernstg: I wanted to remind the reader that we did not previously support the feature in question, as opposed to 'with this update we stopped supporting that feature'.

Another thing is that I do expect some level of support for extracting the actual type arguments "at" some specific type (that is, I expect that we will get some kind of reified existential open), so in that sense it's not unreasonable to read it as "we hope to get it later".

So, trying to come up with something better, I think I'll just step back and leave it as is. ;)

README.md
eernstg: Done!

PR #152, opened 2018-09-14
Fix #132

### LINKED ISSUE #132: reflectable fails with prefixed annotations
```dart
import 'package:dryice/dryice.dart' as di;

@di.injectable
class MaterialApplication {
    void run() {}
}
```
What reflectable builder generates:
```dart
import "dart:core";
import 'dart:async' as prefix10;
import 'dart:html' as prefix9;
import 'package:dryice/dryice.dart' as prefix0;
...

       new r.NonGenericClassMirrorImpl(
            r"MaterialApplication",
            r"mdlcore.MaterialApplication",
            7,
            1,
            const prefix0.InjectAnnotation(),
            const <int>[20, 21],
            const <int>[15, 16, 17, 18, 19, 20],
            const <int>[],
            -1,
            {},
            {},
            {r"": (b) => () => b ? new prefix2.MaterialApplication() : null},
            1,
            1,
            const <int>[],
            const <Object>[prefix0.di.injectable],
            null),
```
This is where the problem occures: `const <Object>[prefix0.di.injectable],`
double prefix...

eernstg:
Reflectable code generation is now stripping client import prefix when generating metadata code for a const variable. For instance, before this PR we generated `prefix2.di.inject`, where `di.inject` is the term used in a client library to access the top level declaration `inject` and `di` is an import prefix, and now we generate something like `prefix2.inject`.

### DIFF
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -4531,7 +4531,7 @@ String _extractMetadataCode(Element element, Resolver resolver,
 String _extractNameWithoutPrefix(Identifier identifier, Element errorTarget) {
   String name;
   if (identifier is SimpleIdentifier) {
-    name = identifier.token.stringValue;
+    name = identifier.token.lexeme;
   } else if (identifier is PrefixedIdentifier) {
     // The identifier is of the form `p.id` where `p` is a library
     // prefix, or it is on the form `C.id` where `C` is a class and
@@ -4540,14 +4540,15 @@ String _extractNameWithoutPrefix(Identifier identifier, Element errorTarget) {
       // We will replace the library prefix by the appropriate prefix for
       // code in the generated library, so we omit the prefix specified in
       // client code.
-      name = "${identifier.identifier.token.stringValue}";
+      name = identifier.identifier.token.lexeme;
     } else {
       // We must preserve the prefix which is a class name.
-      name = "${identifier.name}";
+      name = identifier.name;
     }
   } else {
     _severe("This kind of identifier is not yet supported: $identifier",
         errorTarget);
+    name = identifier.name;
   }
   return name;
 }
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -4534,7 +4534,32 @@ String _extractMetadataCode(Element element, Resolver resolver,
       if (_isPrivateName(annotationNode.name.name)) {
         _severe("Cannot access private name ${annotationNode.name}", element);
       }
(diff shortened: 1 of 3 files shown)


lib/src/builder_implementation.dart
```diff
@@ -4534,13 +4513,43 @@ String _extractMetadataCode(Element element, Resolver resolver,
...
 
   return _formatAsConstList("Object", metadataParts);
 }
 
+/// Extract the plain name from by [identifier] by stripping off the
+/// library import prefix at front, if any.
+String _extractNameWithoutPrefix(Identifier identifier, Element errorTarget) {
+  String name;
+  if (identifier is SimpleIdentifier) {
+    name = "$identifier";
```
sigurdm: I worry about using `toString()` it is usually mostly a debugging tool. Is there not a `Identifier.token` or similar getter you could use returning a string? (Not that it is different from the existing code).

lib/src/builder_implementation.dart
```diff
@@ -4534,13 +4513,43 @@ String _extractMetadataCode(Element element, Resolver resolver,
       if (_isPrivateName(annotationNode.name.name)) {
         _severe("Cannot access private name ${annotationNode.name}", element);
       }
-      metadataParts.add("$prefix${annotationNode.name}");
+      String name = _extractNameWithoutPrefix(annotationNode.name, element);
+      if (_isPrivateName(name)) {
+        _severe("Cannot access private name $name", element);
```
sigurdm: Not specific to this change, but it would be great if we could output a source location of where name is accessed.

lib/src/builder_implementation.dart
```diff
@@ -4534,13 +4513,43 @@ String _extractMetadataCode(Element element, Resolver resolver,
...
+      }
+      metadataParts.add("$prefix$name");
     }
   }
 
   return _formatAsConstList("Object", metadataParts);
 }
 
+/// Extract the plain name from by [identifier] by stripping off the
```
sigurdm: Seems you have a "by" too much

lib/src/builder_implementation.dart
eernstg: This is already in place: We pass `element` as the `errorTarget` which is used to print a source code location, and that gets passed on to `_formatDiagnosticMessage` where the source code location is extracted and included in the returned string.

lib/src/builder_implementation.dart
eernstg: Indeed, fixed!

PR #155, opened 2018-10-26
Port reflectable to build 1.0.x

eernstg:
Flutter depends on `analyzer` 0.33 and that requires a newer `build` (which is now at version 1.0.x), so there's a need to switch to more current versions of several packages in reflectable.

This PR does that; the APIs exported from `build` changed a bit, so the implementation of the builder in `reflectable_builder.dart` had to change somewhat.

The version number has been updated to 2.0.8, such that we can publish this commit directly.

### DIFF
--- a/lib/reflectable_builder.dart
+++ b/lib/reflectable_builder.dart
@@ -8,8 +8,75 @@ import 'dart:async';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
 import 'package:build_config/build_config.dart';
-import 'package:build_runner/build_runner.dart';
+import 'package:build_runner_core/build_runner_core.dart';
+import 'package:build_runner/src/logging/std_io_logging.dart';
+import 'package:build_runner/src/package_graph/build_config_overrides.dart';
+import 'package:build_runner/src/generate/terminator.dart';
 import 'src/builder_implementation.dart';
+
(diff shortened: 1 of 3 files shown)


lib/reflectable_builder.dart
```diff
@@ -50,12 +50,29 @@ Future<BuildResult> reflectableBuild(List<String> arguments) async {
...
-    BuilderOptions options = new BuilderOptions(
+    var options = new BuilderOptions(
         <String, dynamic>{"entry_points": arguments, "formatted": true},
         isRoot: true);
     final builder = new ReflectableBuilder(options);
-    return await build(
-        [applyToRoot(builder, generateFor: new InputSet(include: arguments))],
-        deleteFilesByDefault: true);
+    List<BuilderApplication> builders = [
+      applyToRoot(builder, generateFor: new InputSet(include: arguments))
```
jakemac53: Note that if the input set here changes, that changes the structure of the build, which will cause us to throw away any old work and rebuild the world.

If you modify the builder options instead (which it looks like you already do) then we will only invalidate the builders that use those options.

In this case though, I don't think it really makes a difference since there is only a single builder.

eernstg: OK, then I think I'll leave it as it is for now. Thanks!

PR #158, opened 2019-01-02
Update author for consistency

mit-mit:
Update author to be consistent with all other Dart Team packages


mit-mit:
@kevmoo any objections?

kevmoo:
LGTM!

Issue #159, opened 2019-01-11
Upgrade


manuscle:
Is that possible to upgrade package version dependencies?
When I try to install, I get this error:
The current Dart SDK version is 2.1.0.

```
Because reflectable >=2.0.2 <2.0.8 depends on build ^0.12.0 and reflectable >=2.0.8 depends on analyzer ^0.33.0, reflectable >=2.0.2 requires build ^0.12.0 or analyzer ^0.33.0.
And because angular >=5.2.0 depends on analyzer ^0.34.0 and build_runner >=1.0.0 depends on build >=1.0.0 <1.1.0, one of build_runner >=1.0.0 or reflectable >=2.0.2 or angular >=5.2.0 must be false.
And because reflectable <2.0.2 requires SDK version >=1.8.0 <2.0.0-dev.infinity and agde_mobilite_web depends on angular ^5.2.0, build_runner >=1.0.0 is incompatible with reflectable.
So, because agde_mobilite_web depends on both reflectable any and build_runner ^1.1.2, version solving failed.
```
My pubspec.yaml file:
```
name: my_app
description: A web app that uses AngularDart
version: 0.0.1

environment:
  sdk: '>=2.1.0-dev <3.0.0'

dependencies:
  angular: ^5.2.0
  angular_forms: ^2.0.0
  angular_router: ^2.0.0-alpha+19
  http: ^0.11.0
  stream_transform: ^0.0.6
  angular_components: ^0.10.1
#  sass_builder:
    # This is a temporary work around to have sass_builder include node_modules
    # when it looks for .scss files. The workaround will no longer be needed
    # when https://github.com/dart-league/sass_builder/pull/47 is released.
#    git: https://github.com/jifalops/sass_builder
  sass_builder: ^2.0.2
  intl: ^0.15.7
  cookie_jar: ^0.0.7
  dio:
#  slickdart:
#  mdc_web:
#  ng_bootstrap: ^1.1.1
  bootstrap_sass: ^4.1.3
  crypto:
  uuid: ^1.0.3
  reflectable:

dev_dependencies:
  angular_test: ^2.0.0
  build_runner: ^1.1.2
  build_test: ^0.10.2
  build_web_compilers: ^0.4.0
  test: ^1.0.0
  coverage: ^0.12.3
  mockito: ^3.0.0-beta+1
  pageloader: ^3.0.3
#  intl_translation:
  json_serializable: ^2.0.1
  # grinder is as Dart workflows, automated lib
  grinder:
[...]

PR #160, opened 2019-01-11
Updated analyzer dependency to allow 0.34.1

eernstg:
Cf. issue #159, the version update to allow `0.34.*` is needed for angular >=5.2.0.
This update should make reflectable ready for release as 2.0.9.

This causes a somewhat inconvenient change in the behavior of `build_runner`: It refuses to generate code incrementally (and it writes out a large warning about this). This affects clients (maybe even clients who don't know that they are using reflectable indirectly), and they must add a direct dependency on the analyzer (it is _not_ enough that reflectable has it). I've asked Jacob how to deal with this without asking every direct or indirect client out there to edit their pubspec.yaml and add `analyzer: any`. But I think it makes sense to respond to #159 without having resolved this issue, and then fix it when we know how to do that.

We could use version 2.0.8+2, given that we have now used 2.0.8+1 (and, maybe, those +k versions are the appropriate choice for small things, but it's not obvious to me that a dependency change on such a massive dependency as analyzer is a "small thing"). WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.9
+
+* Updated dependency to allow using analyzer 0.34.
+
 ## 2.0.8+1
 
 * Corrected the email address listed in `pubspec.yaml` for Dart Team.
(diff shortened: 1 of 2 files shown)


eernstg:
OK, thanks! I'll keep 2.0.9 now, and we can still switch to `+k` versions in the future if that turns out to be better.

PR #166, opened 2019-02-25
Pana fixes feb19

eernstg:
The package pana (which is used to compute the 'health' of packages on pub.dartlang.org and similar metrics) is unhappy about a number of things (e.g., `T x = null;` should be `T x;` and `{bool myOptionalParameter: false}` should be `{bool myOptionalParameter = false}`). This PR makes the requested changes.

Note that it includes `// ignore:deprecated_member_use` in the locations where we are using `computeNode` and `unit`. Those problems are real, but it will take a little longer to resolve them. Basically, Brian says that there is more than one user of the AST layer (that is, not the `Element` layer) that the analyzer package has traditionally provided, so there will be ways to keep doing the things that reflectable does via this layer&mdash;but we can't do it right now.

### DIFF
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -296,7 +296,7 @@ class ClassElementEnhancedSet implements Set<ClassElement> {
   bool any(bool f(ClassElement element)) => classElements.items.any(f);
 
   @override
-  List<ClassElement> toList({bool growable: true}) {
+  List<ClassElement> toList({bool growable = true}) {
     return classElements.items.toList(growable: growable);
   }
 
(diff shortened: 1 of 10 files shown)


lib/src/builder_implementation.dart
```diff
@@ -4320,6 +4327,7 @@ String _extractConstantCode(
         Element staticElement = expression.staticElement;
         if (staticElement is PropertyAccessorElement) {
           VariableElement variable = staticElement.variable;
+          // ignore:deprecated_member_use
```
sigurdm: Perhaps add a TODO together with these to indicate that there is more work.

eernstg: Done.

Issue #170, opened 2019-04-19
Unexpected behavior when reflecting on constructor that has parameter that sets instance variable using syntactic sugar


fertrig:
Given the constructor below which uses syntactic sugar to assign a constructor argument to an instance variable.
```
class Sugar {
  Sugar({this.foo = 222});

  final int foo;
}
```
When I invoke the constructor using `newInstance` without any arguments, then the default value of the constructor parameter does not seem to be used. In the example above, after calling `newInstance` without any arguments, the instance variable `foo` is equal to `null`. I'd expect the instance variable `foo` to be equal to `222` after calling `newInstance` without any arguments.

Also, when I get the default value of the constructor parameter, it returns `null`. I'd expect it to return `222`.

I believe the issue is around the syntactic sugar used in the constructor parameter.

I wrote a test to show the failures based on my expectations:
https://gist.github.com/fertrig/8b1c50deb09c6422779fe5014bdd8aeb

Are my expectations correct? Is there a workaround?

eernstg:
Sounds like a bug. (I'm on vacation, but I'll take a look next week).

eernstg:
So quickly a week can evaporate when part of it is vacation. ;-)

The case where a named constructor argument has a default value which is used (because no parameter with that name is passed at a call site) wasn't covered by the current tests. But I just added it ([here](https://github.com/dart-lang/test_reflectable/blob/f635327aa3f56d34c4c8381bcbe7fd68ef4a0705/test/new_instance_default_values_test.dart#L50)), and that test does succeed.

eernstg:
Turns out that an initializing formal constructor parameter (`this.something`) which is optional is _not_ of type `DefaultFormalParameter` in the Dart analyzer program model. So these parameters are not considered when the default value code is generated for the other optional parameters.

I don't know how to obtain that default value at this point (or even how to discover that a given initializing formal parameter has a default value), so I'll have to return to this later.

eernstg:
Awaiting a review, you may be able to use the fix in the branch 'fix_170_may19' by putting this into pubspec.yaml:
```
  reflectable:
    git:
      url: git://github.com/dart-lang/reflectable.git
      ref: fix_170_may19
```
This is expected to be published as 2.0.11.

eernstg:
Closing: The branch is now gone and 2.0.11 has been published.

fertrig:
Thank you so much, @eernstg!

Invoking the constructor now works as expected. 

[...]

PR #171, opened 2019-05-02
Adjusted default value code generation to handle initializing formals

eernstg:
Cf. bug reported as #170.

### DIFF
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -687,29 +687,31 @@ class _ReflectorDomain {
         Iterable.generate(requiredPositionalCount, (int i) => parameterNames[i])
             .join(", ");
 
-    String optionalsWithDefaults =
-        Iterable.generate(optionalPositionalCount, (int i) {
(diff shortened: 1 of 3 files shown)


lib/src/builder_implementation.dart
```diff
@@ -718,8 +726,16 @@ class _ReflectorDomain {
       String defaultValueCode = "";
       if (parameterNode is DefaultFormalParameter &&
           parameterNode.defaultValue != null) {
-        defaultValueCode =
-            ": ${_extractConstantCode(parameterNode.defaultValue, importCollector, _generatedLibraryId, _resolver)}";
+        var constantCode = _extractConstantCode(parameterNode.defaultValue,
```
sigurdm: This is a lot of very similar code - worth extracting a helper?

eernstg: I thought about exactly the same thing, but a quick check made me think they differed here and there. I've checked more precisely now, and moved a 17 line block that occurs twice into a local function.

Issue #173, opened 2019-05-08
Version range for analyzer dependency too small for latest Flutter


suhlatwork:
Similarly to issue #163 (updating to previous stable Flutter), updating a project to the current Flutter stable 1.5.4 fails with unresolvable dependencies.

```
Because reflectable >=2.0.2 <2.0.10 depends on source_span >=1.0.0 <1.5.0 and reflectable >=0.1.1 <2.0.2 requires SDK version >=1.8.0 <2.0.0-dev.infinity, reflectable >=0.1.1 <2.0.10 requires source_span >=1.0.0 <1.5.0.
And because every version of reflected_mustache depends on reflectable ^2.0.0, every version of reflected_mustache requires source_span >=1.0.0 <1.5.0 or reflectable ^2.0.10.
And because reflectable >=2.0.9 depends on analyzer >=0.33.0 <0.35.0 and test_core >=0.2.2 depends on analyzer >=0.35.3 <0.37.0, if reflected_mustache any and test_core >=0.2.2 then source_span >=1.0.0 <1.5.0.
And because test >=1.6.1 <1.6.2 depends on test_core 0.2.3 and test >=1.6.2 depends on test_api 0.2.5, if reflected_mustache any and test >=1.6.1 then source_span >=1.0.0 <1.5.0 or test_api 0.2.5.
Because test >=1.3.0 <1.6.1 depends on stream_channel ^1.6.0 and test <1.3.0 requires SDK version >=1.8.0 <2.0.0-∞, test <1.6.1 requires stream_channel ^1.6.0.
Thus, if reflected_mustache any and test any then source_span >=1.0.0 <1.5.0 or test_api 0.2.5 or stream_channel ^1.6.0.
And because every version of flutter_test from sdk depends on both source_span 1.5.5 and stream_channel 2.0.0, if reflected_mustache any and test any and flutter_test any from sdk then test_api 0.2.5.
And because every version of flutter_test from sdk depends on test_api 0.2.4 and mde_flutter depends on reflected_mustache any, flutter_test from sdk is incompatible with test.
So, because XXX depends on both test any and flutter_test any from sdk, version solving failed.
```

The reason appears to be a discrepancy in the version range of the `analyzer` package (<0.35 by `reflectable`, >=0.35 by `test`/`test_api`).

[...]

PR #179, opened 2019-07-23
Update reflectable to enable analyzer 0.37.0

eernstg:
Cf. issue #173: Code generator updated to use new APIs for access to abstract syntax (`AstNode`) entities (`computeNode` cannot be used any more). This involves a major element of asynchrony, hence there are many changes of functions to be `async`, and of expressions to `await` a result.

The analyzer-as-used-by-build relies on summaries (rather than full libraries) for platform libraries ('dart:...' except 'dart:ui'). This means that it is not possible to obtain information about non-signature properties of platform code. See issue #180 for more details.

Note that the changes in `reflectable_builder.dart` were made to use the control flow shown in `build` package documentation.

This PR introduces a new version number: 2.1.0. This PR is the biggest rewrite of reflectable since 2016 (so it seems misleading to call it 2.0.13), but it is only a breaking change in the sense that clients may not admit the updated versions of various dependencies. This will not actually break clients, they will just stay on 2.0.12 until they are ready to admit the updates&mdash;hence there's no need to go to 3.0.0.

### DIFF
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -708,7 +708,7 @@ class _ReflectorDomain {
       String code = await _extractDefaultValueCode(
           importCollector, constructor.parameters[requiredPositionalCount + i]);
       String defaultPart = code.isEmpty ? "" : " = $code";
-      return "${namedParameterNames[i]}$defaultPart";
+      namedWithDefaultList.add("${namedParameterNames[i]}$defaultPart");
     }
     String namedWithDefaults = namedWithDefaultList.join(", ");
 
@@ -2202,20 +2202,20 @@ class _ReflectorDomain {
       // TODO(eernst): 'dart:*' is not considered valid. To survive, we
       // return the empty metadata for elements from 'dart:*'. Issue 173.
       if (_isPlatformLibrary(element.library)) {
-        return "const []";
-      }
-
-      var resolvedLibrary =
-          await element.session.getResolvedLibraryByElement(element.library);
-      var declaration = resolvedLibrary.getElementDeclaration(element);
-      // The declaration may be null because the element is synthetic, and
-      // then it has no metadata.
-      FormalParameter node = declaration?.node;
-      if (node == null) {
         metadataCode = "const []";
       } else {
-        metadataCode = await _extractMetadataCode(
-            element, _resolver, importCollector, _generatedLibraryId);
+        var resolvedLibrary =
+            await element.session.getResolvedLibraryByElement(element.library);
+        var declaration = resolvedLibrary.getElementDeclaration(element);
+        // The declaration may be null because the element is synthetic, and
+        // then it has no metadata.
+        FormalParameter node = declaration?.node;
+        if (node == null) {
+          metadataCode = "const []";
+        } else {
+          metadataCode = await _extractMetadataCode(
+              element, _resolver, importCollector, _generatedLibraryId);
+        }
       }
     }
     String code = await _extractDefaultValueCode(importCollector, element);
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -152,7 +152,7 @@ class ReflectionWorld {
         typedefs.clear();
       }
       reflectorsCode
-          .add("${reflector._constConstructionCode(importCollector)}: "
+          .add("${await reflector._constConstructionCode(importCollector)}: "
               "$reflectorCode");
     }
     return "final _data = <r.Reflectable, r.ReflectorData>"
@@ -600,7 +600,7 @@ class _ReflectorDomain {
         await _SubtypesFixedPoint(_world.subtypes).expand(_classes);
       }
       if (_capabilities._impliesUpwardsClosure) {
-        await _SuperclassFixedPoint(_capabilities._upwardsClosureBounds,
+        await _SuperclassFixedPoint(await _capabilities._upwardsClosureBounds,
                 _capabilities._impliesMixins)
             .expand(_classes);
       } else {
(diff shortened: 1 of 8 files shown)


lib/src/builder_implementation.dart
```diff
@@ -5125,37 +5261,50 @@ Iterable<DartObject> _getEvaluatedMetadata(
...
+    log.fine(await _formatDiagnosticMessage(message, target));
   } else {
     log.fine(message);
   }
 }
 
 /// Returns a string containing the given [message] and identifying the
 /// associated source code location as the location of the given [target].
-String _formatDiagnosticMessage(String message, Element target) {
+Future<String> _formatDiagnosticMessage(String message, Element target) async {
```
sigurdm: As you suggested offline, it might make things simpler if you gather information for all log messages synchronously and make the async work of getting locations after the main computation.

lib/src/builder_implementation.dart
```diff
@@ -2254,17 +2322,22 @@ class _SuperclassFixedPoint extends FixedPoint<ClassElement> {
   /// class used as a mixin cannot have other superclasses than [Object]).
   /// TODO(eernst) implement: When mixins can have nontrivial superclasses
   /// we may or may not wish to enforce the bounds even for mixins.
-  Iterable<ClassElement> successors(ClassElement element) sync* {
+  Future<Iterable<ClassElement>> successors(ClassElement element) async {
```
sigurdm: It seems to me we could benefit from this method returning a FutureOr.
Only a benchmark could tell how significant the issue is.

lib/src/builder_implementation.dart
```diff
@@ -4830,27 +5019,29 @@ _ClassDomain _createClassDomain(ClassElement type, _ReflectorDomain domain) {
...
+    Element element, AssetId generatedLibraryId, Resolver resolver) async {
+  return await _isImportableLibrary(
+      element.library, generatedLibraryId, resolver);
 }
 
 /// Answers true iff [library] can be imported into [generatedLibraryId].
-bool _isImportableLibrary(
-    LibraryElement library, AssetId generatedLibraryId, Resolver resolver) {
-  Uri importUri = _getImportUri(library, generatedLibraryId);
+Future<bool> _isImportableLibrary(LibraryElement library,
```
sigurdm: Is there any way we can cache the answer for this for all libraries up front?
It seems we would be able to save a lot of async work that way.

pubspec.yaml
```diff
@@ -8,19 +8,19 @@ homepage: https://www.github.com/dart-lang/reflectable
...
+  build_runner: ^1.6.0
+  build_runner_core: ^3.0.0
+  dart_style: '>=1.2.0 <2.0.0'
   glob: ^1.1.0
   logging: ^0.11.0
   package_config: ^1.0.0
   package_resolver: ^1.0.0
-  path: '>=1.2.0 <2.0.0'
-  source_span: ^1.0.0
+  path: '>=1.6.0 <2.0.0'
```
sigurdm: Could use ^ syntax here and a few other places.

lib/src/builder_implementation.dart
```diff
@@ -4251,28 +4341,39 @@ String _formatAsConstList(String typeName, Iterable parts) =>
...
-          expression.elements.map((Expression subExpression) {
-        return helper(subExpression);
-      });
+      List<String> elements = [];
+      for (CollectionElement collectionElement in expression.elements) {
+        if (collectionElement is Expression) {
+          Expression subExpression = collectionElement;
+          elements.add(await helper(subExpression));
+        } else {
+          // TODO(eernst) implement: `if` and `spread` elements of list.
```
sigurdm: Perhaps create an issue to track this

lib/src/builder_implementation.dart
```diff
@@ -131,27 +133,28 @@ class ReflectionWorld {
...
   /// needed to enable the correct behavior for all [reflectors].
-  String generateCode() {
+  Future<String> generateCode() async {
     var typedefs = <FunctionType, int>{};
     var typeVariablesInScope = Set<String>(); // None at top level.
     String typedefsCode = "\n";
-    Iterable<String> reflectorsCode =
-        reflectors.map((_ReflectorDomain reflector) {
+    List<String> reflectorsCode = [];
+    for (_ReflectorDomain reflector in reflectors) {
```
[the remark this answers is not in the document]
eernstg: I'll postpone this one to a later PR.

lib/src/builder_implementation.dart
```diff
@@ -591,13 +594,13 @@ class _ReflectorDomain {
   /// which are directly covered by carrying `_reflector` as metadata or being
   /// matched by a global quantifier, and including the ones which are reached
   /// via the closure operations requested in [_capabilities].
-  ClassElementEnhancedSet get classes {
+  Future<ClassElementEnhancedSet> get classes async {
```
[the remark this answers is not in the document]
eernstg: Done.

lib/src/builder_implementation.dart
```diff
@@ -1988,9 +2034,9 @@ class _ReflectorDomain {
         }
       }
     } else if (dartType is FunctionType) {
-      if (dartType.element.isPrivate) {
-        return 'const r.FakeType(r"${_qualifiedName(dartType.element)}")';
-      }
+      // A function type is inherently not private, so we ignore privacy.
+      // Note that some function types are _claimed_ to be private in analyzer
```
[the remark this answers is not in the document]
eernstg: I have communicated with Brian Wilkerson about this, he thinks it is a bug, and I promised to sort out a minimal example and create an issue.

lib/src/builder_implementation.dart
eernstg: Also covered by https://github.com/dart-lang/reflectable/issues/182.

lib/src/builder_implementation.dart
eernstg: Noted that in #182 as well.

pubspec.yaml
eernstg: Done.

lib/src/builder_implementation.dart
eernstg: That one has been created already (#181).

lib/src/builder_implementation.dart
eernstg: Handled in #182.

PR #184, opened 2019-07-31
Use pedantic lints

eernstg:
Just a small cleanup from enabling pedantic checks: It mostly removes unnecessary `new` and `const` keywords from the code.


pubspec.yaml
```diff
@@ -20,6 +20,7 @@ dependencies:
   package_config: ^1.0.0
   package_resolver: ^1.0.0
   path: ^1.6.0
+  pedantic: ^1.8.0
```
sigurdm: Could this be a dev_dependency?

eernstg: Yes, it still works after that change!

PR #190, opened 2019-12-15
Fix linting issues with generated code

jcowgar:
When using stagehand to generate your base structure, or when including pedantic two linting violations occur with the generated code:

- prefer_single_quotes
- always_declare_return_types

This pull request changes all double quotes to single quotes in generated code and adds the `void` return type to `initializeReflectable()`.

With these changes, `.reflectable.dart` files pass linting. Normally one would not worry so much about generated code, but currently dartanalyzer has a bug where the exclude option is not being processed properly, thus projects can not exclude certain code from analysis. (https://github.com/dart-lang/sdk/issues/29430)


jcowgar:
@googlebot I signed it!

eernstg:
Thanks!

PR #191, opened 2019-12-17
Fix linter issues

eernstg:
This PR changes many parts of the code in order to satisfy new linter requirements. In particular, it uses literal strings with single quotes whenever possible; it omits a number of type annotations on local variables (and introduces `ignore_for_file: omit_local_variable_type` in a couple of libraries, based on the judgment that the value of the type annotations for the reader of the code justifies keeping them).

The generated code now also contains a couple of `ignore_for_file` directives, such that the generated code does not get penalized for repeating structures from the target program (e.g., if a default value is `"a" + " string"` then we will copy that expression rather than simplifying it to `"a string"`, which will cause a linter complaint on the generated code).

Also, an actual bug was detected and fixed: The generation of code dealing with named parameters of constructors (supporting `myClassMirror.newInstance(...)`) relied on two lists having the same order. Turns out that the order may not be the same (at least with the new analyzer), so we are now looking up the name of the parameter at each index.

The resulting package has a new version number, 2.2.1, such that we can publish it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.2.1
+
+* Update reflectable to enable the use of analyzer version 0.39.*. Adjust
+  the implementation to satisfy many new lint requirements, e.g., that
+  string literals use single quotes, sets are created using literals, some
+  local variables have no type annotation, overriding declarations have
+  `@override`, etc. Also fix a bug concerned with the ordering of named
+  parameters in a constructor declaration.
+
 ## 2.2.0
 
 * Adjust implementation to satisfy 'pedantic' lints. The generated code
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,10 @@
 ## 2.2.1
 
-* Update reflectable to enable the use of analyzer version 0.39.*. Adjust
-  the implementation to satisfy many new lint requirements, e.g., that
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.2.1
+
+* Update reflectable to enable the use of analyzer version 0.39.*. Adjust
```
sigurdm: This description is largely irrelevant to the end-user, and covers three parts in one.

Consider starting with the bug fixed, using three bullet-points and limiting the descriptions of the lints
```
* Fix a bug concerned with the ordering of named
  parameters in a constructor declaration.
* Now compatible with `package:analyzer` ^ 0.39.0.
* Generated code is now  lint-free with package:pedantic x.y.z.
* package:reflectable itself  is lint-free with package:pedantic x.y.z.
```

eernstg: That's much better, thanks!

eernstg:
It would indeed have made sense to split this PR in two, based on the nature of the changes. But there is also the perspective where this whole PR arose via "switch to the new analyzer, then repair whatever is now broken".

The bug fix is concerned with lines 722-727 of 'builder_implementation.dart', and nothing else.

Do you wish to go ahead with the split?

eernstg:
Split the bug part out into #192.

PR #193, opened 2019-12-20
Removed author field from pubspec.yaml

eernstg:
Tiny fix, only deletes 'author' field from pubspec.yaml (it is no longer recommended to have this field).


eernstg:
Landing, TBR.

Issue #197, opened 2020-02-25
Code generation exits with error code 1 but is working properly


HerrNiklasRaab:
Hello,

The problem ist, that the command "flutter packages pub run build_runner build" exits with code 1. By removing the "extends Bloc<LoginEvent, LoginState>" from the LoginBloc class the command gets executed without failure. **In both cases, all files are generated properly and the whole project can be run without any problems.** 

In the login_bloc_test.dart the LoginBloc class is referenced but not used.

The LoginBloc class looks like following:

```dart
@inject
class LoginBloc extends Bloc<LoginEvent, LoginState> {
  UserService userService;

  @inject
  LoginBloc(this.userService);

  @override
  Stream<LoginState> mapEventToState(LoginEvent event) async* {
  }

  @override
  LoginState get initialState => LoginState();
}
```

Executing "flutter packages pub run build_runner build -v" results in the following output:

```
niklasraab@Niklass-MBP fazua_toolbox_app % flutter packages pub run build_runner build -v
[INFO] Generating build script...
[INFO] Generating build script completed, took 423ms

[INFO] BuildDefinition:Initializing inputs
[INFO] BuildDefinition:Reading cached asset graph...
[INFO] BuildDefinition:Reading cached asset graph completed, took 96ms

[INFO] BuildDefinition:Checking for updates since last build...
[INFO] BuildDefinition:Checking for updates since last build completed, took 1.0s

[INFO] Build:Running build...
[INFO] Heartbeat:1.0s elapsed, 0/1 actions completed.
[INFO] Heartbeat:5.8s elapsed, 0/1 actions completed.
[INFO] Heartbeat:12.4s elapsed, 0/1 actions completed.
[SEVERE] reflectable:reflectable on test/blocs/login_bloc_test.dart:
`reflectedTypeArguments` where an actual type argument (possibly nested) is void
[INFO] Heartbeat:14.4s elapsed, 0/1 actions completed.
[WARNING] reflectable:reflectable on test/blocs/login_bloc_test.dart:
Attempt to generate code for an unsupported kind of type: void (VoidTypeImpl). Generating `dynamic`.
[WARNING] reflectable:reflectable on test/blocs/login_bloc_test.dart:
Attempt to generate code for an unsupported kind of type: void (VoidTypeImpl). Generating `dynamic`.
[WARNING] reflectable:reflectable on test/blocs/login_bloc_test.dart:
[...]

PR #199, opened 2020-02-27
Adding support for a reflected type reifying the type `void`

eernstg:
The type `void` used to have no representation at run time, but Dart was updated a while back such that `void` is now reified. This PR changes the treatment of `void` as a `reflectedType` such that this reification is made available. Cf. #197.


eernstg:
About the lower SDK bound: No, we have lower bound 2.3.0 and `void` as a reified type works at least as far back as 2.1.0-dev.0.0.

Issue #210, opened 2020-06-11
Build broken on recent Flutter dev channel versions


RiccardoM:
I've always developed my Flutter app using the `dev` channel of the Flutter SDK. This has recently resulted in the following `reflectable` build error: 

```
type '() => (_SyncIterator<PropertyAccessorElement>) => bool' is not a subtype of type '() => (_SyncIterator<ExecutableElement>) => bool
```

I've tried many different Flutter SDK versions, and here are the results: 

- :ballot_box_with_check: `1.19.0-1.0.pre` or previous
- :warning: `1.19.0-2.0.pre` or later

eernstg:
Here's branch that seems to resolve the issue: https://github.com/dart-lang/reflectable/tree/use_newer_analyzer_jun20.

RiccardoM:
I can confirm this is fixed on such branch. The code to use it (or anyone that might end up here in the future) currently is: 

```yaml
dependencies:
  reflectable:
    git:
      url: https://github.com/dart-lang/reflectable
      ref: use_newer_analyzer_jun20
```

eernstg:
@RiccardoM, thanks for trying it out!

The branch `use_newer_analyzer_jun20` has been merged and published as reflectable 2.2.5, so I'll delete the branch.

PR #211, opened 2020-06-15
Update analyzer; work around sync* getter issue

eernstg:
Cf. #210 and https://github.com/dart-lang/sdk/issues/42347, there is an issue with `sync*` getters, This PR works around that issue, and upgrades the analyzer version to 0.39.10.

Note that the "workaround" is adding `.toList()` at a point where several other similar expressions already have `.toList()`, which seems to be a reasonable approach anyway, and probably a bit faster. So there is no need to undo it when dart-lang/sdk#42347 is handled.

### DIFF
--- a/lib/src/builder_implementation.dart
+++ b/lib/src/builder_implementation.dart
@@ -923,7 +923,7 @@ class _ReflectorDomain {
       var classesToAdd = <ClassElement>{};
       ClassElement anyClassElement;
       for (ClassElement classElement in await classes) {
-        if (_typeForReflectable(classElement).isObject) {
+        if (_typeForReflectable(classElement).isDartCoreObject) {
           hasObject = true;
           objectClassElement = classElement;
           break;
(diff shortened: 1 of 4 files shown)


lib/src/builder_implementation.dart
```diff
@@ -2758,13 +2758,11 @@ class _ClassDomain {
...
   /// the same semantics as that of `declarations` in [ClassMirror].
-  Iterable<ExecutableElement> get _declarations {
+  Iterable<ExecutableElement> get _declarations sync* {
     // TODO(sigurdm) feature: Include type variables (if we keep them).
-    return () sync* {
-      yield* _declaredMethods;
-      yield* _accessors;
-      yield* _constructors;
-    }();
+    yield* _declaredMethods;
```
sigurdm, line 2763: If it is not important that we construct this lazily we could just use:
```
[
  ..._declaredMethods,
  ..._acessors,
  ..._constructors,
]
```

eernstg: I think we're iterating over these collections in full and several times, so it's quite likely to be faster to compute lists. Done, including a few other `sync*` functions.

PR #212, opened 2020-07-02
Prepare for mono_repo

eernstg:
This PR moves the current contents of the repository into a new subdirectory 'reflectable', adds another subdirectory 'test_reflectable' and copies the current contents of the repository 'test_reflectable' into that subdirectory (omitting a couple of tests which were demonstrating a limitation that we do not have a solution for), preparing the repository for a setup using `mono_repo`.


eernstg:
Thanks! I'll enable travis in a later step, and drop the history.

PR #217, opened 2020-08-29
Widen versioning constraint to analyzer to >=0.39.4 <0.40.0

Fox32:
I had some trouble with build runner after doing a pub upgrade, in the end I noticed that some unwanted versions were used because of the analyzer version constraint in reflectable. So I increased it, I haven't observed any issues with it. But I don't know if the limit was intentional.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.2.5+1
+
+* Widen versioning constraint to analyzer to >=0.39.4 <0.40.0
+
 ## 2.2.5
 
 * Update dependencies. In particular, analyzer can now be 0.39.10.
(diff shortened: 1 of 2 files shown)


gwxtr:
+1
ideally upgrade analyzer dependency to latest one (0.40.2 at the moment)

Fox32:
Yeah, that has to do someone else, I had the feeling that >=0.40.0 contains bigger changes

eernstg:
We can't promise to work with an unlimited number of versions of the analyzer, because the combination of reflectable with every single version of the analyzer needs to be verified (this is because reflectable relies on certain parts of the analyzer which are not officially supported). But https://github.com/google/reflectable.dart/pull/220 broadens the constraint to all the versions 0.39.* that have been published so far, and that should be the complete list since there are several versions 0.40.* today. They are deprecating a large number of features, so it will be a somewhat more delicate step to use a 0.40.* analyzer.

Fox32:
@eernstg thanks for solving this!

eernstg:
Thank you!

Issue #198, opened 2020-02-27
Code generation causes InconsistentAnalysisException with multiple targets


eernstg:
The following commands show that reflectable currently fails to generate code when dependencies have been fully updated (`pub upgrade`):

```
> git clone 'https://github.com/dart-lang/test_reflectable.git'
> cd test_reflectable
> cd tool
> make
```

This causes a number of failures like this:

```
[SEVERE] reflectable:reflectable on test/invoke_test.dart.vm_test.dart:

InconsistentAnalysisException: Requested result might be inconsistent with previously returned results
```

However, it only seems to affect code generation in the case where there are multiple targets, that is, when code generation occurs for multiple separate entry points ("programs").

We do not yet know why this happens, but it may be caused by some changes in the way the build related packages give notifications to the analyzer about changed files, cf. https://github.com/dart-lang/build/issues/2634.

eernstg:
[This PR](https://github.com/google/reflectable.dart/pull/220) uses the workaround from dart-lang/build#2634 to eliminate the inconsistency.

eernstg:
Finally closing this one: Resolved by #220!

PR #220, opened 2020-09-11
Use workaround for 198

eernstg:
Issue #198 reports an `InconsistentAnalysisException` when code generation is performed with more than one entry point. A workaround came up in dart-lang/build#2634, and this PR uses that workaround, which does eliminate the dynamic errors. Prepared to publish as version 2.2.6.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## 2.2.6
 
 * Resolve #198 (InconsistentAnalysisException), based on workaround from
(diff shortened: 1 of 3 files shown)


sigurdm (approved):
LGTM

Is there also a regression test for this?

eernstg:
With this change, code generation for tests are again working (before the change, code could only be generated for one test at a time). There is no test which will run code generation, but this will be part of the change where we run the code generation via Travis. A PR for that is coming.

PR #224, opened 2020-10-07
Fix constant evaluator invocation

eernstg:
Unfortunately, the published version 2.2.7 allows analyzer 0.40.3, but that version has a breaking change (affecting `ConstantEvaluator(..., ... /*This argument*/)`), so 2.2.7 only works with analyzers up to 0.40.2. This PR prepares publishing 2.2.8 which adapts the code to work after said change, and supports analyzer 0.40.3 - 0.40.4.

### DIFF
--- a/test_reflectable/pubspec.yaml
+++ b/test_reflectable/pubspec.yaml
@@ -10,11 +10,11 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
 dependencies:
   build_runner: any
-  glob: ^1.2.0
+  glob: any
   reflectable:
     path: ../reflectable
 dev_dependencies:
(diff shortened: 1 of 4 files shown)


test_reflectable/pubspec.yaml
```diff
@@ -10,11 +10,11 @@ environment:
   sdk: '>=2.3.0 <3.0.0'
 dependencies:
   build_runner: any
-  glob: ^1.2.0
+  glob: any
```
eernstg, line 13: test_reflectable now uses `any` for all dependencies, such that all version constraints are taken from reflectable itself.

stefantsov (approved):

Issue #227, opened 2020-11-07
[reflectable 2.2.8] doesn't create main.reflectable


aliyazdi75:
After upgrading my flutter, reflectable doesn't create `main.reflectable.dart`.
I don't know what causes this problem but I share my logs.
<details>
  <summary>flutter packages pub run build_runner build:</summary>
  
```
[INFO] Generating build script...
[INFO] Generating build script completed, took 484ms

[INFO] Initializing inputs
[INFO] Reading cached asset graph...
[INFO] Reading cached asset graph completed, took 99ms

[INFO] Checking for updates since last build...
[INFO] Checking for updates since last build completed, took 1.2s

[INFO] Running build...
[WARNING] reflectable:reflectable on test/widget_test.dart:
Your current `analyzer` version may not fully support your current SDK version.

Please try upgrading to the latest `analyzer` by running `flutter packages upgrade`.

Analyzer language version: 2.11.0
SDK language version: 2.12.0

If you are getting this message and have the latest analyzer please file
an issue at https://github.com/dart-lang/sdk/issues/new with the title
"No published analyzer available for language version 2.12.0".
Please search the issue tracker first and thumbs up and/or subscribe to
existing issues if present to avoid duplicates.

[INFO] 1.4s elapsed, 0/1 actions completed.
[INFO] 2.6s elapsed, 0/1 actions completed.
[INFO] 3.7s elapsed, 0/1 actions completed.
[INFO] 4.8s elapsed, 0/1 actions completed.
[INFO] 10.4s elapsed, 0/1 actions completed.
[INFO] 21.2s elapsed, 0/1 actions completed.
[WARNING] No actions completed for 21.2s, waiting on:
  - reflectable:reflectable on test/widget_test.dart

[INFO] 22.3s elapsed, 0/1 actions completed.
[INFO] 23.3s elapsed, 0/1 actions completed.
[INFO] 24.3s elapsed, 0/1 actions completed.
[INFO] 25.4s elapsed, 0/1 actions completed.
[INFO] 26.4s elapsed, 0/1 actions completed.
[INFO] 27.5s elapsed, 0/1 actions completed.
[INFO] 28.5s elapsed, 0/1 actions completed.
[INFO] 29.6s elapsed, 0/1 actions completed.
[INFO] Running build completed, took 30.4s

[INFO] Caching finalized dependency graph...
[INFO] Caching finalized dependency graph completed, took 55ms

[...]

PR #228, opened 2020-12-01
Update analyzer version and fix 198, 227

eernstg:
Update analyzer dependency to allow all versions until breaking change in 0.41.0. Issue #227 reports that 'lib/main.dart' is skipped during code generation; this PR fixes that. Issue #198 about `InconsistentAnalysisException` re-emerged with analyzer 0.40.5, workaround from https://github.com/google/built_value.dart/issues/941 adopted.

This PR also adds `// @dart=2.9` to tests, because they do not get the language version from the pubspec.yaml.

New patchset uploaded Dec 3rd: Package `build` was updated and analyzer 0.41.0, 0.41.1 are now working. Changed pubspec.yaml and CHANGELOG.md accordingly.

### DIFF
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -5430,8 +5430,19 @@ Future<ResolvedLibraryResult> _getResolvedLibrary(
   // mentioned in dart-lang/build#2634. If we do not fetch a fresh session
   // then the code generation stops with an `InconsistentAnalysisException`
   // if there is more than one entry point.
-  final freshLibrary =
-      await resolver.libraryFor(await resolver.assetIdForElement(library));
-  final freshSession = freshLibrary.session;
-  return await freshSession.getResolvedLibraryByElement(freshLibrary);
+  var attempts = 0;
+  while (true) {
+    try {
+      final freshLibrary =
+          await resolver.libraryFor(await resolver.assetIdForElement(library));
+      final freshSession = freshLibrary.session;
+      return await freshSession.getResolvedLibraryByElement(freshLibrary);
+    } catch (_) {
+      ++attempts;
+      if (attempts == 10) {
+        log.severe('Analysis session did not stabilize after ten tries!');
+        return null;
+      }
+    }
+  }
 }
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -5429,7 +5429,11 @@ Future<ResolvedLibraryResult> _getResolvedLibrary(
   // This is expensive, but seems to be necessary. It is using the workaround
   // mentioned in dart-lang/build#2634. If we do not fetch a fresh session
   // then the code generation stops with an `InconsistentAnalysisException`
-  // if there is more than one entry point.
+  // if there is more than one entry point. This workaround turned out to be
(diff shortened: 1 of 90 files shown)


reflectable/lib/src/builder_implementation.dart
```diff
@@ -5430,8 +5430,19 @@ Future<ResolvedLibraryResult> _getResolvedLibrary(
...
+  while (true) {
+    try {
+      final freshLibrary =
+          await resolver.libraryFor(await resolver.assetIdForElement(library));
+      final freshSession = freshLibrary.session;
+      return await freshSession.getResolvedLibraryByElement(freshLibrary);
+    } catch (_) {
+      ++attempts;
+      if (attempts == 10) {
+        log.severe('Analysis session did not stabilize after ten tries!');
```
sigurdm, line 5443: This error message is not really understandable by the user. I don't know if we can give more context or better explain what is going on.

Is there anything the user can do here - should they just try again?

Is this really an internal error?

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5430,8 +5430,19 @@ Future<ResolvedLibraryResult> _getResolvedLibrary(
   // mentioned in dart-lang/build#2634. If we do not fetch a fresh session
   // then the code generation stops with an `InconsistentAnalysisException`
   // if there is more than one entry point.
-  final freshLibrary =
-      await resolver.libraryFor(await resolver.assetIdForElement(library));
-  final freshSession = freshLibrary.session;
-  return await freshSession.getResolvedLibraryByElement(freshLibrary);
+  var attempts = 0;
```
sigurdm, line 5433: How sad that this is necessary.

I'm not sure I understand the underlying problem fully. But at least we should have a reference to where this trick was borrowed from:
```suggestion
  // This hack was suggested in https://github.com/google/built_value.dart/issues/941#issuecomment-731077220.
  var attempts = 0;
```

sigurdm (approved):
LGTM

reflectable/lib/src/builder_implementation.dart
eernstg: Issue #198 was rather confusing for a long time, and then suddenly jakemac53 contributed a workaround [here](https://github.com/dart-lang/build/issues/2634#issuecomment-670603224), with no explanation about how the issue arises in the first place. "Someone calls `AnalysisDriver.changeFile()`", but it is not at all obvious why that would happen for a library which is part of a given entry point during reflectable code generation, even in the case where multiple entry points from `test_reflectable/test` have their code generation performed in an interleaved fashion: The only generated library for an entry point `<e>.dart` is `<e>.reflectable.dart`, so there is no overlap in generated libraries for two distinct tests.

So it's still largely unexplained, and users probably can't do anything other than trying again (except that I've never seen those 10 attempts exhausted, so it might never happen).

> Is this really an internal error?

I think it is a bug in `build_runner`, and in that sense it would be internal to the reflectable code generator, except that it could also be caused by `while true; do touch <e>.dart; done`.

Do you prefer to change the error message in order to reflect the situation more precisely?

reflectable/lib/src/builder_implementation.dart
eernstg: Done.

PR #231, opened 2021-01-09
Generate code with nullsafety

eernstg:
This PR migrates the support libraries (used by a running program) to enable null safety, and updates the code generation libraries (esp. `lib/src/builder_implementation.dart`) to be able to generate code that enables null safety (it uses `// @dart=2.9` when the entry point does not enable null safety).

It also creates a variant of each test in test_reflectable that uses `@dart=2.9` to stay at the old language version as well as the variant of the test which is migrated to enable null safety.

The command `pub run build_runner test` does not work on the latter, because it insists on running them using `--sound-null-safety`, which fails because `build_test` isn't yet null safe. This issue may be resolved, but it is possible to run the test manually (like `dart --no-sound-null-safety test/basic_test.dart`).

Just one test is failing (`declaration_test.dart`, line 150), and that's a known issue.


reflectable/lib/capability.dart
```diff
@@ -502,13 +502,13 @@ class _StringInvocation extends StringInvocation {
   final Map<Symbol, dynamic> namedArguments;
 
   @override
-  bool isMethod;
+  late bool isMethod;
 
   @override
-  bool isGetter;
+  late bool isGetter;
```
sigurdm, line 508: Not sure where we initialize these. Should they not be getters that inspect `kind`?

reflectable/lib/capability.dart
```diff
@@ -518,11 +518,11 @@ class _StringInvocation extends StringInvocation {
 /// doesn't have the capabilities to invoke it.
 class ReflectableNoSuchMethodError extends Error
     implements NoSuchCapabilityError {
-  final Object receiver;
+  final Object? receiver;
```
sigurdm, line 521: Could we add a line of documentation explaining what it means for this to be null?

reflectable/lib/capability.dart
```diff
@@ -518,11 +518,11 @@ class _StringInvocation extends StringInvocation {
...
   final String memberName;
   final List positionalArguments;
-  final Map<Symbol, dynamic> namedArguments;
-  final List<Symbol> existingArgumentNames;
+  final Map<Symbol, dynamic>? namedArguments;
```
sigurdm, line 524: Could we add a line of documentation explaining what it means for this to be null? Why not just an empty map?

reflectable/lib/capability.dart
```diff
@@ -518,11 +518,11 @@ class _StringInvocation extends StringInvocation {
...
+  final List<Symbol>? existingArgumentNames;
```
sigurdm, line 525: Ditto here

reflectable/lib/mirrors.dart
```diff
@@ -489,7 +489,7 @@ abstract class InstanceMirror implements ObjectMirror {
   ///     }
   ///
   /// Required capabilities: [delegate] requires the [delegateCapability].
-  dynamic delegate(Invocation invocation);
+  Object? delegate(Invocation invocation);
```
sigurdm, line 492: There's a typo in the comment above: `invocation.positionArguments[0]);` should be `invocation.positionalArguments[0]);`

sigurdm, line 492: Is it a breaking change to return `Object?` instead of `dynamic`? Should it be mentioned in the changelog?

reflectable/lib/reflectable.dart
```diff
@@ -4,8 +4,6 @@
 
 // This is the main library in package reflectable.
```
sigurdm, line 6: Did this TODO get fixed?

reflectable/lib/reflectable.dart
```diff
@@ -4,8 +4,6 @@
...
```
sigurdm, line 5: This comment doesn't provide much. Perhaps we should delete it?

reflectable/lib/reflectable.dart
```diff
@@ -39,10 +37,10 @@ abstract class ReflectableInterface {
   /// it can be obtained from `originalDeclaration`. That mirror will
   /// have no actual type arguments, but it will have declared type
   /// variables. Other types than classes are not (yet) supported.
-  TypeMirror reflectType(Type type);
+  TypeMirror? reflectType(Type type);
```
sigurdm, line 40: Can we document when this might return null?

reflectable/lib/reflectable.dart
```diff
@@ -39,10 +37,10 @@ abstract class ReflectableInterface {
...
 
   /// Returns a mirror of the given library [library].
-  LibraryMirror findLibrary(String library);
+  LibraryMirror? findLibrary(String library);
```
sigurdm, line 43: Can we document when this might return null?

reflectable/lib/src/builder_implementation.dart
```diff
@@ -4109,10 +4118,14 @@ class BuilderImplementation {
     }
     imports.sort();
 
+    String languageVersionComment =
+        world.entryPointLibrary.isNonNullableByDefault
+            ? ''
```
sigurdm, line 4123: Should we add a `@dart = 2.12` to this case to future proof?

sigurdm (approved):
Nice
Mostly just suggestions to document the null-return value and a question about some late fields.

reflectable/lib/capability.dart
eernstg: Definitely yes, thanks!

reflectable/lib/capability.dart
eernstg: That was indeed useful: A static method uses `reflectedType` as the receiver, but that would throw when no capability was requested to have that `reflectedType`, so I changed it to test `hasReflectedType` and use null if not. The remaining cases look ok.

reflectable/lib/capability.dart
eernstg: The implementation does handle a `namedArguments` whose value is null, and the only occurrences of null aware invocations (like `namedArguments?.keys`) existed already before this PR. Hence, allowing clients of reflectable to pass null for this argument is an existing feature. I don't see the need to take that feature away.

On the other hand, `_StringInvocation` does replace that null by an empty map, so the error messages won't talk about null. Added a dartdoc comment.

reflectable/lib/capability.dart
eernstg: Deleted the field and associated parameters: Turns out that no call sites are able to provide this information, and it was always null.

Clients are not expected to create instances of `ReflectableNoSuchMethodError`, nor are they expected to call `reflectableNoSuch{Method,{G,S}etter,Constructor}Error`. So I jumped to the conclusion that we don't need to make this parameter deprecated, and removed it.

reflectable/lib/mirrors.dart
eernstg: Thx!

eernstg: All methods named `invoke`/`invokeGetter`/`invokeSetter` use the return type `Object?` and `newInstance` uses `Object`, so `delegate` was the only one returning `dynamic`. Added the change to the change log. It is breaking, but the fix is local, so I think we should be OK.

reflectable/lib/reflectable.dart
eernstg: Right, done.

reflectable/lib/reflectable.dart
eernstg: Done.

reflectable/lib/reflectable.dart
eernstg: It can't. Changed the return type to non-nullable.

reflectable/lib/reflectable.dart
eernstg: It can't. Changed the return type to non-nullable.

reflectable/lib/src/builder_implementation.dart
eernstg: I'm not sure: If someone is using old tools then language version 2.12 may not exist (but they may have been using `--enable-experiment=non-nullable`, and we don't have to break them).

The other direction should be safe: If we specify language version 2.12 and the tools and client package are at 2.17 then it _ought_ to be working (because it takes version ^3.0.0-0 to break), but the introduction of null safety shows that it's nowhere near safe to jump silently into a version 2.x.y.

So I added it.

eernstg:
Thanks for the very useful input! I responded to all threads, fixed various stuff, resolved all but two. The two threads that are still open are concerned with breaking changes: I've chosen to go ahead and perform the breaking change.

For `existingArgumentNames`: I think nobody should pass this argument, or even call the function that declares this parameter, so it's not unreasonable to go ahead.

For `delegate`: It is definitely a breaking change to change the return type from `dynamic` to `Object?`. However, it is surprising that this is the only method returning `dynamic` (related methods return `Object?`: `invoke`/`invokeGetter`/etc.).

So WDYT, do we need to have more of a breaking change process?

sigurdm:
In my opinion these breaks are alright without much more process - switching to null-safety should be a breaking change anyway - just important to document them as much as possible.

I would also not spend too much energy in supporting old tools running with `--enable-experiment=non-nullable` - we must assume people using experiments are interested in upgrading as soon as possible  - otherwise they can always stay on the earlier version of reflectable until they are able to upgrade.

eernstg:
Right, thanks!

PR #233, opened 2021-01-20
Use `astNodeFor`, and avoid generation of unused code

eernstg:
A small cleanup PR:

Allow the latest analyzer 0.42.2.

`Resolver.astNodeFor` is a new method that allows us to avoid the (expensive, slow) workaround needed to avoid `InconsistentAnalysisException`; this PR uses `astNodeFor` where possible.

The input IDs of the form `*.vm_test.*`, `*.node_test.*`, `*.browser_test.*` give rise to code generation where the generated code is never used (they are never imported); this PR skips code generation for these input IDs. This improves the speed of code generation, e.g., `test_reflectable` code generation is reduced on my desktop from `114.41s user 10.92s system 164% cpu 1:16.04 total` to `69.81s user 8.40s system 166% cpu 46.990 total` (a speedup factor of 1.64).

It is prepared for publishing as 3.0.0-nullsafety.1.

Latest patchset: Require *-nullsafety version of test in test_reflectable/pubspec.yaml; turns out that it is not impossible to get it, but it is not enough to have `test: ^1.16.0-nullsafety` in reflectable/pubspec.yaml, the client _must_ request the non-stable version itself. This allows us to avoid all those "wrong migration order" hints, because we now avoid importing a legacy library from an opted-in library. Also, we can now run with implicit null checking (so the legacy version of each test will run as with `--no-sound-null-safety` and the opted-in version will run as with `--sound-null-checking`)!


sigurdm (approved):
LGTM

eernstg:
Thx!

reflectable/pubspec.yaml
```diff
@@ -1,15 +1,15 @@
...
+version: 3.0.0-nullsafety.1
 description: >
   Reflection support based on code generation, using 'capabilities' to
   specify which operations to support, on which objects.
 homepage: https://github.com/google/reflectable.dart
 environment:
   sdk: '>=2.12.0-0 <3.0.0'
 dependencies:
-  analyzer: '>=0.41.0 <=0.41.1'
+  analyzer: '>=0.41.0 <=0.41.2'
```
sigurdm, line 10: Should this not require at least 0.41.2 to use the new api?

sigurdm: Ah scrap that. The new api is in package:build

PR #237, opened 2021-03-15
Create stable version of reflectable with null safety

eernstg:
Prepare a release as 3.0.0 (not 3.0.0-nullsafety.X), that is, a stable release with null safety support.

Includes changes around `FunctionTypeAliasElement` (changed is part of the non-function type aliases feature work), changes to use `type.element` rather than `type` when checking subclass relationships (some new occurrences of types of the form `T*` made the subclass test fail in some cases where the subclass relationship did actually hold), and finally updated a lot of versions to stable rather than `-nullsafety.X`.


sigurdm (approved):
LGTM

eernstg:
Thx!

cedvdb:
Pub dev doesn't seem to label this package as supporting null safety on https://pub.dev/packages/reflectable

eernstg:
Right, that's a spurious complaint: Reflectable contains some libraries which are intended to be imported by other libraries, and other libraries which are the implementation of the code generator.

It does not make any client package less null safe that the code _generator_ hasn't yet been fully migrated, but the pub scoring mechanism can't see which libraries are intended to be imported by client packages and which ones are part of the code generator (or any other thing that creates a similar situation).

We can't go further right now:

```
...> pub outdated --mode=null-safety
Showing dependencies that are currently not opted in to null-safety.
[✗] indicates versions without null safety support.
[✓] indicates versions opting in to null safety.

Package Name       Current  Upgradable  Resolvable  Latest   

direct dependencies:
build_config       ✗0.4.7   ✗0.4.7      ✗0.4.7      ✗0.4.7   
build_runner       ✗1.12.1  ✗1.12.1     ✗1.12.1     ✗1.12.1  
build_runner_core  ✗6.1.12  ✗6.1.12     ✗6.1.12     ✗6.1.12  
dart_style         ✗1.3.14  ✗1.3.14     ✗1.3.14     ✗1.3.14  

dev_dependencies: all support null safety.
You are already using the newest resolvable versions listed in the 'Resolvable' column.
Newer versions, listed in 'Latest', may not be mutually compatible.
```

PR #248, opened 2021-04-27
Use nullsafe build, migrate to null safety

eernstg:
This PR migrates the reflectable code generator to null safety, and it uses the newly arrived null safe versions of build packages, such that reflectable is fully opted in to null safety.

In general, nulls that arise because the program does not pass static analysis are asserted (so there are lots of ways that the code generation will crash with a null error if we run it on a program with errors). We may wish to change this later on, to avoid all those possible crashes.

However, it does not seem useful to replicate a large portion of error messages about Dart errors in any code generator, because it's just a duplicate of something that we already have in a much more complete form in the analyzer and the CFE. 

The best approach would probably be to ask the analyzer on startup whether this program has any errors, and only generate code if not.

### DIFF
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -1259,12 +1259,12 @@ class _ReflectorDomain {
     return result;
   }
 
-  Future<int> _computeOwnerIndex(
+  Future<int?> _computeOwnerIndex(
       ExecutableElement element, int descriptor) async {
     if (element.enclosingElement is ClassElement) {
-      return (await classes).indexOf(element.enclosingElement)!;
+      return (await classes).indexOf(element.enclosingElement);
     } else if (element.enclosingElement is CompilationUnitElement) {
-      return _libraries.indexOf(element.enclosingElement.enclosingElement!)!;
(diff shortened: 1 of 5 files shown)


reflectable/lib/src/builder_implementation.dart
```diff
@@ -4371,24 +4428,24 @@ int _declarationDescriptor(ExecutableElement element) {
   }
 
   if (element is PropertyAccessorElement) {
-    result = element.isGetter ? constants.getter : constants.setter;
+    result |= element.isGetter ? constants.getter : constants.setter;
     handleReturnType(element);
   } else if (element is ConstructorElement) {
     if (element.isFactory) {
-      result = constants.factoryConstructor;
+      result |= constants.factoryConstructor;
```
eernstg, line 4435: It seems less error-prone to initialize `result` to zero and use `|=` everywhere, rather than (as previously) keeping an eye on which bit-setting operations occur first in the control flow, and using `=` on exactly those bit-setting operations.

stefantsov (approved):

buzzware:
Yay! Thanks to all involved. I just got back to this :) @eernstg

Issue #253, opened 2021-07-20
Error: "Null check operator used on a null value" while using `reflectedTypeCapability`


salah-rashad:
Hi, first of all, _I would like to apologize for my bad English, it's not my native language_.

So, I was trying to reflect on a Class that extends from an abstract class, everything is okay when I'm NOT using `reflectedTypeCapability`, but when I do then run the build runner again I get this error:

> Null check operator used on a null value

- I tried `flutter pub run build_runner clean`, the same error shows up.

<details>
  <summary>console output</summary>

```
PS F:\PROJECTS\flamer> flutter pub run build_runner build -v
[  +80 ms] executing: [C:\flutter/] git -c log.showSignature=false log -n 1 --pretty=format:%H
[  +64 ms] Exit code 0 from: git -c log.showSignature=false log -n 1 --pretty=format:%H
[   +1 ms] f4abaa0735eba4dfd8f33f73363911d63931fe03
[   +1 ms] executing: [C:\flutter/] git tag --points-at f4abaa0735eba4dfd8f33f73363911d63931fe03
[  +50 ms] Exit code 0 from: git tag --points-at f4abaa0735eba4dfd8f33f73363911d63931fe03
[        ] 2.2.3
[   +8 ms] executing: [C:\flutter/] git rev-parse --abbrev-ref --symbolic @{u}
[  +76 ms] Exit code 0 from: git rev-parse --abbrev-ref --symbolic @{u}
[   +1 ms] origin/stable
[   +1 ms] executing: [C:\flutter/] git ls-remote --get-url origin
[  +44 ms] Exit code 0 from: git ls-remote --get-url origin
[   +1 ms] https://github.com/flutter/flutter.git
[ +151 ms] executing: [C:\flutter/] git rev-parse --abbrev-ref HEAD
[  +42 ms] Exit code 0 from: git rev-parse --abbrev-ref HEAD
[        ] stable
[ +113 ms] Artifact Instance of 'AndroidGenSnapshotArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'AndroidInternalBuildArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'IOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[   +3 ms] Artifact Instance of 'WindowsEngineArtifacts' is not required, skipping update.
[...]

PR #254, opened 2021-07-29
Fix a null error; update dependencies

eernstg:
Remove a `!` and handle the null which may actually occur (fixing #253). Change the analyzer dependency to ^2.0.0, and perform a handful of changes that follow from this.

One test was changed (presumably a bug was fixed in the analyzer): An initializing formal parameter is now reported to be final, and a test expectation has been changed to reflect this new behavior.


stefantsov (approved):

eernstg:
Thanks, @stefantsov!

PR #268, opened 2022-03-16
Fix enum constructor and stray null

### LINKED ISSUE #267: Fails to generate compilable code
```
lib/main.reflectable.dart:821:54: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'deprecated', 1082517, null,
                                                     ^
lib/main.reflectable.dart:823:52: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'override', 1082517, null,
                                                   ^
lib/main.reflectable.dart:825:55: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'provisional', 1082517, null,
                                                      ^
lib/main.reflectable.dart:827:49: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'proxy', 1082517, null, const prefix0.Reflector(),
                                                ^
lib/main.reflectable.dart:829:56: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'keepToString', 1082517, null,
                                                       ^
lib/main.reflectable.dart:831:58: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'channelBuffers', 1082389, null,
                                                         ^
lib/main.reflectable.dart:833:50: Error: The value 'null' can't be assigned to the parameter type 'int' because 'int' is not nullable.
        r.VariableMirrorImpl(r'window', 1082389, null,
                                                 ^
```

```dart
        r.VariableMirrorImpl(r'deprecated', 1082517, null,
            const prefix0.Reflector(), -1, 19, 19, const <int>[], const []),
```

Based on everything else, that 3rd parameter should be an int index, not sure why its not the case.

eernstg:
This PR fixes #267 (where null was generated where an expression of type `int` is required), and a non-reported bug where a constructor invocation could be generated for an `enum` type. It prepares release 3.0.9 as well.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.0.8
+
+* Change code generation such that no invocations of generative constructors
+  of enums are generated (such that it works with 'enhanced-enums').
+
 ## 3.0.7
 
 * Change `analysis_options.yaml` such that we get the same linting locally
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.0.9
+
+* Correct bug reported as #267. Correct code generation which is reported
+  as an error now that we have enhanced enums.
+
 ## 3.0.8
 
 * Change code generation such that no invocations of generative constructors
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -1682,7 +1682,8 @@ class _ReflectorDomain {
       bool reflectedTypeRequested) async {
     int descriptor = _topLevelVariableDescriptor(element);
     var owner = element.library;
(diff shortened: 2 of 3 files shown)


eernstg:
(Oops, I updated CHANGELOG twice, introducing version 3.0.8 as well as 3.0.9. I've corrected that now, so we only have a section on 3.0.8, and pubspec.yaml also says 3.0.8. The 3.0.9 thing just happened because it has been several days since I wrote the bits about 3.0.8, and I forgot ;-)

chloestefantsova (approved):

Issue #278, opened 2022-06-30
async feature?


explodus:
how to invoke and wait on async methods? isn't it implemented yet?

eernstg:
There would not be anything special about that: You would invoke a method using reflection rather than directly, and you would receive the returned result. That result would be a future `f`, and you'd await it using `await f` as usual.

explodus:
> There would not be anything special about that: You would invoke a method using reflection rather than directly, and you would receive the returned result. That result would be a future `f`, and you'd await it using `await f` as usual.

thanks, as i wrote this lines, i saw it by myself. could i ask another question?

`
class DartReflectable extends r.Reflectable {
  const DartReflectable()
      : super(
          invokingCapability,
          newInstanceCapability,
          declarationsCapability,
          reflectedTypeCapability,
          typingCapability,
          subtypeQuantifyCapability,
          superclassQuantifyCapability,
          typeAnnotationQuantifyCapability,
        );
}

@RestApi(baseUrl: 'https://rest.somewhere.com/')
@DartReflectable()
abstract class MyClient {
  ///
  factory MyClient(Dio dio, {String baseUrl}) = _MyClient;
}
`

it's a retrofit generated class, i got all things prepared/generated, but a can't get a new instance from it, cause it has no informations about _MyClient. so i can't invoke any method and can only print informations

eernstg:
You should be able to obtain a class mirror `cm` on `MyClient`, and then `cm.newInstance('', [someDio], {#baseUrl:'someUrl'})` should yield an instance of `_MyClient`, because the constructor in `MyClient` is a factory constructor that actually redirects the call to the constructor in `_MyClient`.

However, you can't get a mirror for `_MyClient`, because reflectable cannot generate any code that would denote that class (you can't cheat on privacy in Dart). This means that you still can't call methods on an instance of `_MyClient` reflectively. I guess the declaration of `_MyClient` is generated, so you don't have the option to give it a non-private name?

explodus:
no it doesn't work. i tried it also with generating an extra class like 

```
@DartReflectable()
class MyClientReflectable extends _MyClient {
  MyClientReflectable(Dio dio, {String? baseUrl}) : super(dio, baseUrl: baseUrl);
}
```

to get the constructor getting called/invoked 

here are some of the error messages
```
         var reflectable = const DartReflectable();
         final myClass = reflectable.reflectType(MyClient) as ClassMirror;
         instance = reflectable.reflect(
[...]

PR #281, opened 2022-07-11
Added required_inputs directive to build.yaml

eernstg:
Cf. https://github.com/google/reflectable.dart/issues/278#issuecomment-1179381155.

This PR changes the 'build.yaml' of reflectable such that it has `required_inputs: [".dart"]`. This causes the build runner to schedule the reflectable code generator as the last code generator (among those who are generating any `*.dart` files, at least, but that's all we need). The consequence is that the reflectable code generator gets to see all the code generated by other code generators. This is needed in order to generate reflection support for all the applicable parts of the given program (rather than skipping some of the declarations generated by other code generators).

This PR also prepares reflectable for release as 3.0.10.


reflectable/build.yaml
```diff
@@ -4,6 +4,7 @@ builders:
     import: "package:reflectable/reflectable_builder.dart"
     builder_factories: ["reflectableBuilder"]
     build_extensions: {".dart": [".reflectable.dart"]}
+    required_inputs: [".dart"]
```
natebosch, line 7: @jakemac53 - double checking that you've had a chance to see this pattern and don't have any concerns with it.

Reflectable seems like it makes sense to both require and output .dart. Only 1 such builder can exist in a given project, but I think that is fine.

jakemac53: I am a bit conflicted about it - but as a practical matter it makes sense for this package.

It does mean  you won't be able to use this package together with another builder which also has the same basic constraints though (but they might anyways be fundamentally incompatible).

eernstg: If it is a choice that usually works, could a client package (a package that depends on reflectable and runs the code generator) override that choice in their own build.yaml? In that case it could serve as a default setting that (presumably) almost always works.

eernstg: Otherwise, https://github.com/dart-lang/build/pull/3334 looks like the right thing to do.

jakemac53: They can't easily override it (technically it is possible, you can override the entire build.yaml from package:reflectable, but it is not advised to do so and is a maintenance burden).

I am not opposed to this change in principal, it has both pros and cons, but I would probably just tell users to use the new feature coming in https://github.com/dart-lang/build/pull/3334 to avoid future issues.

natebosch: > It does mean you won't be able to use this package together with another builder which also has the same basic constraints though (but they might anyways be fundamentally incompatible).

I had the same concerns and ended up leaning towards the assumption that they'd be fundamentally incompatible anyway.

> I am not opposed to this change in principal, it has both pros and cons, but I would probably just tell users to use the new feature coming in [dart-lang/build#3334](https://github.com/dart-lang/build/pull/3334) to avoid future issues.

It's hard for me to get a sense of whether this current minor pain is likely to hit enough users to override the theoretical major pain if the incompatibility comes up. If we've only had this come up for a single user, then the extra config is the right tradeoff. Does anyone have any idea if there are many other users with the same problem?

jakemac53: > I had the same concerns and ended up leaning towards the assumption that they'd be fundamentally incompatible anyway.

Fwiw, I don't actually think that is a fair assumption to make. The type of code that reflectable outputs is unlikely to be needed by other "global" builders. It could likely always run last.

eernstg:
The safest approach would be to rely on https://github.com/dart-lang/build/pull/3334, which is a feature that allows clients to specify code generator ordering. This PR will be abandoned if that approach works, because the approach used in this PR will create a problem for clients who want to use two "compiling code generators", which could surely occur in real life.

eernstg:
Closing this PR: https://github.com/dart-lang/build/pull/3334 seems to be the best solution.

reflectable/build.yaml
eernstg: > It could likely always run last.

I'll keep that in mind if the issue comes up again.

PR #283, opened 2022-07-20
Add support for generic type arguments in annotations

jacksonjude:
# Overview
Adds support for generic annotations, a feature added in [Dart SDK 2.14](https://github.com/dart-lang/language/blob/master/accepted/2.14/small-features-21Q1/feature-specification.md).

Take this class for example:
```dart
@reflector
class Person {
  @TypeHelper<Soul>(42, "The meaning")
  String? life;
}

class TypeHelper<T> {
  const TypeHelper(int var1, String var2);
}

class Soul { }
```
The annotation would have previously been identified as having the type argument `dynamic`:
```dart
ClassMirror personClassMirror = reflector.reflectType(Person) as ClassMirror;
VariableMirror field = personClassMirror.declarations["life"] as VariableMirror;
var annotation = field.metadata[0];
print(annotation.runtimeType); // TypeHelper<dynamic>
```
It will now be correctly identified as having the type argument `Soul`:
```dart
print(annotation.runtimeType); // TypeHelper<Soul>
```

This is done by appending the type arguments (if any exist) after the annotation name in during the `_extractMetadataCode` function of `builder_implementation.dart`.

### DIFF
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -4877,7 +4877,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
       if (_isPrivateName(name)) {
         await _severe('Cannot access private name $name', element);
       }
-      metadataParts.add('const $prefix$name($arguments)');
+      if (annotationNode.typeArguments != null) {
+        String typeArguments = annotationNode.typeArguments!.arguments
+            .map((TypeAnnotation typeArgument) {
+              LibraryElement library = typeArgument.type!.element!.library!;
+              String prefix = importCollector._getPrefix(library);
+              return '$prefix$typeArgument';
+            })
+            .join(', ');
+        metadataParts.add('const $prefix$name<$typeArguments>($arguments)');
+      } else {
+        metadataParts.add('const $prefix$name($arguments)');
+      }
     } else {
       // A field reference.
       if (_isPrivateName(annotationNode.name.name)) {


eernstg:
Hello @jacksonjude, thanks for the PR! This is something reflectable should support. I'm really busy now, but I'll take a look as soon as possible.

jacksonjude:
Hey there, sorry if you are still busy, but can you take another look at this?

eernstg:
Sorry about the delay &mdash; yep, I'm still busy, but I will take a look at it tomorrow.

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5005,7 +5005,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
       if (_isPrivateName(name)) {
         await _severe('Cannot access private name $name', element);
       }
-      metadataParts.add('const $prefix$name($arguments)');
+      if (annotationNode.typeArguments != null) {
+        String typeArguments = annotationNode.typeArguments!.arguments
```
eernstg, line 5009: So this becomes:
```suggestion
        String typeArguments = typeArguments.arguments
```

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5005,7 +5005,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
       if (_isPrivateName(name)) {
         await _severe('Cannot access private name $name', element);
       }
-      metadataParts.add('const $prefix$name($arguments)');
+      if (annotationNode.typeArguments != null) {
```
eernstg, line 5008: We usually use a style (maybe it isn't consistent in this library, but we're trying to be consistent at least with code that gets updated ;-) where null is handled. So we'd have something like this:

```suggestion
      var typeArguments = annotationNode.typeArguments;
      if (typeArguments != null) {
```

This means that the local variable `typeArguments` is promoted to have type `TypeArgumentList` rather than `TypeArgumentList?`, and then we don't have to test for null when it's used later.

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5005,7 +5005,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
...
+        String typeArguments = annotationNode.typeArguments!.arguments
+            .map((TypeAnnotation typeArgument) {
+              LibraryElement library = typeArgument.type!.element!.library!;
```
eernstg, line 5011: We can't assume that all those getters return a non-null value. In particular, a small test that I just wrote makes the code generator crash in this line.

```
Null check operator used on a null value
package:reflectable/src/builder_implementation.dart 5011:71  _extractMetadataCode.<fn>
```

It looks like it is `arguments` (which ends at column 71).

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5005,7 +5005,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
...
+              String prefix = importCollector._getPrefix(library);
```
eernstg, line 5012: For the prefix we need to check that the given library is importable (`_isImportable`). For instance, `dart:_internal` is not importable by user code.

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5005,7 +5005,18 @@ Future<String> _extractMetadataCode(Element element, Resolver resolver,
...
+              return '$prefix$typeArgument';
```
eernstg, line 5013: I think this won't work in the case where the type argument is a parameterized type of the form `C<D>` where `C` is imported from `prefix1` and `D` is imported from `prefix2`: We should get `prefix1.C<prefix2.D>`, but the code above simply uses `toString()` on the given `TypeAnnotation`, and that will yield syntax which is only valid for the location where that type annotation actually occurs in the program (and at this point we are writing some code that should denote the same type, but in a different library).

Check out `_typeCodeOfTypeArgument`, you might be able to call that function in order to get the right string.

eernstg (changes requested):
Thanks again for the input! We need to make the code a bit more robust. Here is the beginnings of a test that is intended to verify that the new feature is working:

```dart
@c
library test_reflectable.test.metadata_test;

import 'package:reflectable/reflectable.dart';
import 'package:test/test.dart';
import 'metadata_test.reflectable.dart';

class MyReflectable extends Reflectable {
  const MyReflectable()
      : super(metadataCapability, instanceInvokeCapability,
            staticInvokeCapability, declarationsCapability, libraryCapability);
}

const myReflectable = MyReflectable();

const c = [
  Bar<num>({'a': 14})
];

class K {
  static const p = 2;
}

@myReflectable
@Bar<Object?>({
  b: deprecated,
  c: Deprecated('tomorrow'),
  1 + 2: (d ? 3 : 4),
  identical(1, 2): 's',
  K.p: 6
})
@Bar<Never>({})
@Bar<void>.namedConstructor({})
@c
class Foo {
  @Bar<bool>({})
  @Bar<Bar<Bar>>({})
  @Bar<MyReflectable>.namedConstructor({})
  @Bar<void Function()>.namedConstructor({})
  @c
  void foo() {}
  var x = 10;
}

class Bar<X> {
  final Map<Object, X> m;
  const Bar(this.m);
  const Bar.namedConstructor(this.m);

  @override
  String toString() => 'Bar<$X>($m)';
}

void main() {
  initializeReflectable();

  test('metadata on class', () {
    expect(myReflectable.reflectType(Foo).metadata, const [
      MyReflectable(),
      Bar<Object?>({
        b: deprecated,
        c: Deprecated('tomorrow'),
        3: 3,
        false: 's',
        2: 6,
      }),
      [
        Bar<num>({'a': 14})
      ],
    ]);

    var fooMirror = myReflectable.reflectType(Foo) as ClassMirror;
    expect(fooMirror.declarations['foo']!.metadata, const [
      Bar<bool>({}),
      Bar<Bar<Bar<dynamic>>>({}),
      Bar<MyReflectable>({}),
      Bar<void Function()>({}),
      [
        Bar<num>({'a': 14})
      ],
    ]);

    // The synthetic accessors do not have metadata.
    expect(fooMirror.instanceMembers['x']!.metadata, []);
    expect(fooMirror.instanceMembers['x=']!.metadata, []);

    // Test metadata on libraries
    expect(myReflectable.reflectType(Foo).owner!.metadata, [c]);
  });
}
```

The test probably needs to be corrected here and there (but it already shows that the new code gets a null reference error, so it's a useful starting point).

The test should be added to the repository at `test_reflectable/test/generic_metadata_test.dart`. The lower bound of the SDK constraint in test_reflectable/pubspec.yaml` should be changed to 2.17 (such that the generic-metadata feature is included in the language).

reflectable/lib/src/builder_implementation.dart
jacksonjude: Having some trouble calling `_typeCodeOfTypeArgument` since it is a function of `_ReflectorDomain`, which I can't seem to access from within `_extractMetadataCode`. However, it is very possible I may have missed something given that I'm not super familiar with the layout of the library.

reflectable/lib/src/builder_implementation.dart
jacksonjude: I think I've resolved this, I currently have the type argument default to `'dynamic'` if the library can't be imported, though I'm not sure if this is the correct behavior.

reflectable/lib/src/builder_implementation.dart
jacksonjude: Added some checks for `typeArgument.type` and `typeArgument.type.element`, and that seemed to fix the error

jacksonjude:
One more question: the variables `b` and `d` don't seem to be defined in the sample test code you have provided. Is this intentional?

eernstg:
(No, I haven't forgotten this. Just busy again. Will try to get to it real soon now. ;-)

eernstg:
(Busy, busy, busy, but I will take a look later today.)

codekeyz:
@eernstg

Sten435:
@codekeyz @eernstg Any updates on this ?

It has been a year ?

jacksonjude:
Hi again @eernstg, what changes need to be made for this PR to be ready to merge?

eernstg:
Hi @jacksonjude, thanks for contributing! I'm sincerely sorry about not being able to get around to this for so long.

I took another look at it today, and I can see that it generates code where a prefix is missing when this test is executed:

```dart
@c
library test_reflectable.test.generic_metadata_test;

import 'package:reflectable/reflectable.dart';
import 'package:test/test.dart';
import 'metadata_test.reflectable.dart';

class MyReflectable extends Reflectable {
  const MyReflectable()
      : super(metadataCapability, instanceInvokeCapability,
            staticInvokeCapability, declarationsCapability, libraryCapability);
}

const myReflectable = MyReflectable();

const b = 0;

const c = [
  Bar<num>({'a': 14})
];

const d = true;

class K {
  static const p = 2;
}

@myReflectable
@Bar<Object?>({
  b: deprecated,
  c: Deprecated('tomorrow'),
  1 + 2: (d ? 3 : 4),
  identical(1, 2): 's',
  K.p: 6
})
@Bar<Never>({})
@Bar<void>.namedConstructor({})
@c
class Foo {
  @Bar<bool>({})
  @Bar<Bar<Bar>>({})
  @Bar<MyReflectable>.namedConstructor({})
  @Bar<void Function()>.namedConstructor({})
  @c
  void foo() {}
  var x = 10;
}

class Bar<X> {
  final Map<Object, X> m;
  const Bar(this.m);
  const Bar.namedConstructor(this.m);

  @override
  String toString() => 'Bar<$X>($m)';
}

void main() {
  initializeReflectable();

  test('metadata on class', () {
    expect(myReflectable.reflectType(Foo).metadata, const [
      MyReflectable(),
      Bar<Object?>({
        b: deprecated,
        c: Deprecated('tomorrow'),
        3: 3,
        false: 's',
        2: 6,
      }),
      [
        Bar<num>({'a': 14})
      ],
    ]);

    var fooMirror = myReflectable.reflectType(Foo) as ClassMirror;
    expect(fooMirror.declarations['foo']!.metadata, const [
      Bar<bool>({}),
      Bar<Bar<Bar<dynamic>>>({}),
      Bar<MyReflectable>({}),
      Bar<void Function()>({}),
      [
        Bar<num>({'a': 14})
      ],
    ]);

    // The synthetic accessors do not have metadata.
    expect(fooMirror.instanceMembers['x']!.metadata, []);
    expect(fooMirror.instanceMembers['x=']!.metadata, []);

    // Test metadata on libraries
    expect(myReflectable.reflectType(Foo).owner!.metadata, [c]);
  });
}
```

The error from the analyzer looks as follows:

```
  error • test/generic_metadata_test.reflectable.dart:66:45 • The name 'Bar' isn't a type, so it can't be used as a type argument. Try correcting the name to an existing type, or defining a type named 'Bar'. • non_type_as_type_argument
```

If I edit the generated code to add the missing prefix then I get 'Null check operator used on a null value'.

I think it shouldn't be hard to fix. It would be great to get it landed!

jacksonjude:
Hi again @eernstg, I just patched the bug relating to nested type arguments. The remaining null errors seem to be the result of the `b` and `d` variables not being defined within `generic_metadata_test.dart`.

jacksonjude:
Hey @eernstg, it turns out I was using an older version of the test script you sent, and the new one (I believe) had a typo at the top: (`import 'metadata_test.reflectable.dart';` instead of `import 'generic_metadata_test.reflectable.dart';`. Changing that should resolve the null error you were initially getting.

On another note, I did realize I was repeating a lot of the code from `_ReflectorDomain._typeCodeOfTypeArgument`, so I did some work to make that method static (in order for me to call it within `_extractMetadataCode`). All of the other tests still run correctly after these changes.

Onto the metadata test itself, I am a little confused on why the first expect (`myReflectable.reflectType(Foo).metadata`) would return 3 elements, seeing as there are 5 annotations attached to `Foo`. Should `@Bar<Never>({})` and `@Bar<void>.namedConstructor({})` explicitly not be included because they contain invalid type arguments?

PR #287, opened 2022-08-31
Upated-with-analyzer-4.6.0

ashutosh2014:
Fix https://github.com/google/reflectable.dart/issues/286

PR #288, opened 2022-09-12
Update analyzer to 5.0.0

eernstg:
The analyzer changed in several ways, and this migration gave rise to many changes. In particular, mixins and enums are no longer represented as `ClassElement`s with an `isEnum` and an `isMixin`, and this causes many changes to types and type tests.

### DIFF
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -4409,7 +4409,7 @@ int _classDescriptor(InterfaceElement element) {
   int result = constants.clazz;
   if (element.isPrivate) result |= constants.privateAttribute;
   if (element.isSynthetic) result |= constants.syntheticAttribute;
-  if (element is MixinElement && element.isAbstract ||
+  if (element is MixinElement ||
       element is ClassElement && element.isAbstract) {
     result |= constants.abstractAttribute;
   }
@@ -5487,7 +5487,7 @@ class MixinApplication implements ClassElement {
   bool get isAbstract {
     var mixin = this.mixin;
     return !isMixinApplication ||
-        mixin is MixinElement && mixin.isAbstract ||
+        mixin is MixinElement ||
         mixin is ClassElement && mixin.isAbstract;
   }
 
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -225,7 +225,7 @@ class Enumerator<T extends Object> {
 /// Hybrid data structure holding [InterfaceElement]s and supporting data.
 /// It holds three different data structures used to describe the set of
 /// classes under transformation. It holds an [Enumerator] named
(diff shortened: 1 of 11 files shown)


gmpassos:
👍🏻 👏🏻 just waiting for the release! Thanx

eernstg:
#287, #289 will be released first, but this one is still needed because we have to stop using a bunch of APIs that are deprecated (and indeed deleted from analyzer 5.0.0).

eernstg:
@sigurdm, this one is rather long (a whole bunch of methods have been deprecated in the analyzer). Would it be possible for you to find the time to review it?

reflectable/lib/src/builder_implementation.dart
```diff
@@ -2774,8 +2846,8 @@ class _ClassDomain {
 
   /// Finds all instance members by going through the class hierarchy.
   Iterable<ExecutableElement> get _instanceMembers {
-    Map<String, ExecutableElement> helper(ClassElement classElement) {
-      var member = _reflectorDomain._instanceMemberCache[classElement];
+    Map<String, ExecutableElement> helper(InterfaceElement interfaceElement) {
+      var member = _reflectorDomain._instanceMemberCache[interfaceElement];
```
sigurdm, line 2850: Maybe this could be written more clearly with
```suggestion
      return _reflectorDomain._instanceMemberCache.putIfAbsent(interfaceElement, () {
      ....
      
   }
```

reflectable/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
 name: reflectable
-version: 3.0.10
+version: 3.1.0
```
sigurdm, line 2: Consider doing a major increment now that you have breaking changes.

sigurdm (approved):
LGTM.

reflectable/lib/src/builder_implementation.dart
```diff
@@ -215,34 +222,35 @@ class Enumerator<T extends Object> {
   }
 }
 
-/// Hybrid data structure holding [ClassElement]s and supporting data.
+/// Hybrid data structure holding [InterfaceElement]s and supporting data.
 /// It holds three different data structures used to describe the set of
 /// classes under transformation. It holds an [Enumerator] named
-/// [classElements] which as a set-like data structure that also enables
+/// [interfaceElements] which as a set-like data structure that also enables
```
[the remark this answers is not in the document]
eernstg: Indeed, fixed by changing `as` to `is`.

reflectable/pubspec.yaml
eernstg: Yes! For quite a while I was working on this PR targeting analyzer 4.4.0 (and that one already contained the breaking change on the analyzer side about mixins), but analyzer 5.0.0 came out since then, and this means that we can piggyback on the new major version of the analyzer. So 4.0.0 it is.

PR #302, opened 2022-12-05
Remove language version 2.12

eernstg:
The reflectable code generator generates the language version comment `// @dart = 2.9` when the program uses unsound null safety (that is: at least one library is not null safe). It used to use `// @dart = 2.12` in programs using sound null safety. However, that's harmful because it prevents usage of more recent language features in the generated code. (In particular, constant expressions may do this, because they are built from expressions in libraries that are part of the target program.)

This PR removes the language version comments from every generated library which would have had `// @dart = 2.12`, which means that it will use the default language version of the package that it is generated into.

The PR prepares reflectable for release as 4.0.3 as well.


sigurdm (approved):
LGTM

eernstg:
Thanks!

Issue #305, opened 2022-12-12
Cannot reflect ```flutter/material.dart``` URIs for ```Container``` type


pattobrien:
Hi all -

I'm creating an ```analyzer``` based utility package, to more-easily allow developers to create lint rules, quick assists, etc, and one of the individual proof-of-concepts I'm designing is a more intuitive API for interacting with AST utilities like type-checkers. For more context, you can check out [sidecar](https://github.com/pattobrien/sidecar) and [reflectable_flutter_analyzer](https://github.com/pattobrien/reflectable_flutter_analyzer).

As the title explains, I'm unable to reflect ```flutter/material.dart``` type URIs, which are needed to perform runtime-checking of a URI match between an analyzer Element's URI and the expected URI of the Type (e.g. Container). I have this working for ```dart.ui.Color``` but it does not work for ```Container``` from ```package:flutter```:

```dart
import 'package:flutter/material.dart';

import 'package:reflectable_analyzer/reflectable_analyzer.dart';

@GlobalQuantifyCapability(r'\.Color$', analyzerReflector)
@GlobalQuantifyCapability(r'\.Container$', analyzerReflector)
import 'package:reflectable/reflectable.dart' hide SourceLocation;

import 'reflector.reflectable.dart';

void main() {
  initializeReflectable();
}
```

The above code incorrectly generates the following:

```dart
...
      <m.LibraryMirror>[
        r.LibraryMirrorImpl( 
            r'dart.ui', // Correct library mirror for ```dart.ui.Color```
            Uri.parse(r'reflectable://0/dart.ui'),
            ...
        ),
        r.LibraryMirrorImpl(
            r'',  // Blank library mirror for ```flutter.material.Container```
            Uri.parse(r'reflectable://1/'),
            ...
        ),
      ],
...
```

```dart
 <m.TypeMirror>[
        r.NonGenericClassMirrorImpl(
            r'Color',
            r'dart.ui.Color', // Correct qualified name
            ...
        ),
       r.NonGenericClassMirrorImpl(
            r'Container',
            r'.Container', // Expected ```flutter.material.Container``` or something similar
            ...
       ),
```

eernstg:
I suspect that 'package:flutter/material.dart' is treated as a special case by the analyzer, and hence there is no access to the declarations in that library.

I'm not sure why that would be necessary, but other platform libraries behave in a similar manner.

I can't immediately experiment with the situation as described:

```
> flutter pub get
[...]

PR #310, opened 2023-02-13
Upgrade to prepare for Dart 3; use working uris when available

eernstg:
This PR deletes the legacy tests (they contain non-null-safe code and cannot be executed in the future). The PR also adjusts the SDK bounds in order to prepare for Dart 3.

The update revealed that one test, 'test_reflectable/test/new_instance_native_test.dart', performed a reflective invocation of the list constructor named `List`, which is no longer supported. The test was adjusted to call another constructor of `List`.

Finally, this PR changes the computation of the `uri` of a `LibraryMirror` such that it uses a `package` URI whenever possible. This should be helpful in the situation where there is a need to find the actual source code, as requested in https://github.com/google/reflectable.dart/issues/305.

The change in the result returned by the `uri` getter may seem like a breaking change, but it is hardly breaking in practice because the result returned previously was essentially meaningless (using the scheme `reflectable` which is not used anywhere else, and consisting mostly of a number which reflected the numbering of libraries being loaded by the analyzer during code generation).

The new behavior is tested in the updated version of 'reflect_type_test.dart'.


reflectable/lib/src/builder_implementation.dart
```diff
@@ -2234,12 +2234,25 @@ class _ReflectorDomain {
...
+    String uriCode;
+    if (_capabilities._supportsUri || _capabilities._supportsLibraries) {
+      AssetId? assetId;
+      try {
+        assetId = await _resolver.assetIdForElement(library);
+      } catch (_) {
+        assetId = null;
+      }
+      if (assetId != null) {
+        var uri = assetId.uri.replace(scheme: 'package');
```
sigurdm, line 2248: Reading https://pub.dev/documentation/build/latest/build/AssetId/uri.html I would expect this to already be a package-uri for practical purposes. And if not a package-uri I'm not sure we want to rewrite the scheme (maybe just replace by some placeholder...).

test_reflectable/test/reflect_type_test.dart
```diff
@@ -56,6 +56,8 @@ void main() {
     var libraryMirror =
         broadReflector.findLibrary('test_reflectable.test.reflect_type_test');
     expect(libraryMirror, isNotNull);
-    expect(libraryMirror.uri.toString().contains('reflect_type_test'), true);
+    var uriString = libraryMirror.uri.toString();
+    expect(uriString.contains('package:'), true);
```
sigurdm, line 60: Maybe we should expect the exact uri here.

sigurdm (approved):
LGTM with a comment about the Urls...

reflectable/lib/src/builder_implementation.dart
eernstg: Interesting, I should have looked that up. ;-)

I noticed that the URI obtained using `AssetId.uri` is `asset:test_reflectable/test/reflect_type_test.dart` for the test 'reflect_type_test.dart' in the package `test_reflectable`.

The 'uri.replace(scheme: 'package')' actually won't work as a package uri an Dart source code. So maybe we should simply keep the 'asset' scheme, and leave it to users of this feature to look up how to interpret an `asset` URI.

test_reflectable/test/reflect_type_test.dart
eernstg: Doing that now.

reflectable/lib/src/builder_implementation.dart
sigurdm: That sounds good.

I guess there is no package-uri for anything outside lib/ so it kind-of makes sense (and makes writing the tests a bit harder).

It would be good to leave a comment about how these Uris are generated (perhaps with a pointer to https://pub.dev/documentation/build/latest/build/Resolver/assetIdForElement.html), that would make it easier to consume.

Also the documentation could mention that package-uris can be looked up via: https://api.dart.dev/stable/2.19.2/dart-isolate/Isolate/packageConfig.html .

sigurdm: Or maybe better: https://api.dart.dev/stable/2.19.2/dart-isolate/Isolate/resolvePackageUri.html

eernstg: Thanks! -- added those links to CHANGELOG.md.

sigurdm: I would expect it to go here: https://pub.dev/documentation/reflectable/latest/reflectable.mirrors/LibraryMirror/uri.html

PR #326, opened 2024-01-05
Upgrade reflectable to use new version of analyzer

eernstg:
This PR changes reflectable such that it uses the analyzer version 6.4.0, and also updates other dependencies.

This includes several changes to the code associated with evaluation of constant expressions (where the new version of the analyzer does not support the approach which was used for this purpose previously).

### DIFF
--- a/reflectable/lib/src/builder_implementation.dart
+++ b/reflectable/lib/src/builder_implementation.dart
@@ -7,6 +7,7 @@ library reflectable.src.builder_implementation;
 // ignore_for_file:implementation_imports
 
 import 'dart:developer' as developer;
+import 'package:analyzer/dart/analysis/declared_variables.dart';
 import 'package:analyzer/dart/analysis/session.dart';
 import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/ast/ast.dart';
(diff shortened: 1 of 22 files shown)


matt-hall-zory:
It looks like you have path ^1.9.0 set as the minimum version.
However, this will rule out use of integration_test which is pinned by flutter to ^1.8.3

eernstg:
Thanks for the heads-up, @matt-hall-zory! I'll adjust the version to 1.8.3.

PS: I'm waiting for a feature that would restore the ability to evaluate constants using analyzer 6.3.0, so I can't land this one right now.

Sten435 (approved):

eernstg:
We still don't have "official" support for constant evaluation in the analyzer, but I've taken a shortcut: Importing some `src` libraries. This is not guaranteed to continue to work (because breaking changes in `src` libraries are not breaking ;-), but it will get us on track again.

@sigurdm, would you be able to take a look at this PR?

stan-at-work (approved):

eernstg:
@stan-at-work and @Sten435, thanks for the reviewing votes!

For this package (or anything in github.com/google and similar locations), I do need an approving review from a Google employee as well. So that's the reason why this hasn't landed yet.

reflectable/lib/src/builder_implementation.dart
```diff
@@ -5582,10 +5584,66 @@ bool _isPrivateName(String name) {
   return name.startsWith('_') || name.contains('._');
 }
 
-EvaluationResult _evaluateConstant(
-    LibraryElement library, Expression expression) {
-  return ConstantEvaluator((library as LibraryElementImpl).source, library)
-      .evaluate(expression);
+DartObject? _evaluateConstant(LibraryElement library, Expression expression) {
+  // !!!TODO!!! Analyzing without errors until this point.
```
sigurdm, line 5588: Is this something that needs to be addressed before landing?

Otherwise we should probably file an issue and format this as
```suggestion
  // TODO(https://github.com/google/reflectable/issues/<issue number>) Analyzing without errors until this point.
```

sigurdm (approved):

reflectable/lib/src/builder_implementation.dart
eernstg: Yes, certainly, `!!!TODO!!!` should never be committed. Thanks for catching that!

eernstg: Changed the function to `async`, now using `_severe` like everywhere else. No more TODOs. ;-)

eernstg:
:tada: Unstuck!

codekeyz:
Amazing work @eernstg

PR #328, opened 2024-07-02
Update reflectable dependencies to use SDK 3.4.0 and analyzer 6.7.0

eernstg:
Update the package dependencies and resolve static analysis issues caused by this change.

In particular, remove several occurrences of named parameters that are now deprecated (e.g., `withNullability`), change invocations of `variable` to `variable2` (the analyzer package made this change in order to be able to migrate `variable` to a new semantics).

The new getter `variable2` has a nullable return type. It can only be null in the case where the given declaration is augmenting. Reflectable uses a defensive approach to handle these nulls, and it will most likely need to be updated when the augmentation feature is fully specified and implemented.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.7
+
+* Update dependencies to use sdk 3.4.0 and analyzer 6.7.0.
+
 ## 4.0.6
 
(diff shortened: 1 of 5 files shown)


sigurdm (approved):
LGTM!

eernstg:
Thanks!

PR #332, opened 2024-08-22
Update dependencies

eernstg:
This PR updates reflectable to use analyzer ^6.7.0. This was previously not possible (because using analyzer 6.7.0 caused version resolution to fail), but it seems to work at this time. It is prepared for releasing version 4.0.10 of reflectable.


sigurdm (approved):

carasen12:
https://github.com/google/reflectable.dart/pull/332#issuecomment-2304612143

PR #335, opened 2024-10-23
Revert deps to analyzer 6.8.0

eernstg:
The dependency upgrades in 4.0.11 caused no issues locally, but on github they cause failures of basically all the score related checks (pub version resolution fails with respect to some macro packages). This PR changes the dependency on the analyzer to 6.8.0 and lints to 5.0.0, and reverts some changes that were required with analyzer 6.11.0 but which cause errors with analyzer 6.8.0.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 4.0.12
 
-* Reverting analyzer dependency to 6.8.0 due to macros conflict.
+* Revert analyzer dependency to 6.8.0 and lints to 5.0.0 due to macro
+  related version resolution conflict.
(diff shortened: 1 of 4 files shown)


eernstg:
@sigurdm, the newest analyzer is incompatible with the version of the tools that are running score checks on pub.dev. This PR downgrades the versions in such a way that a stable tool chain (3.5.4) locally has no version resolution failures.

I'll land this PR now and upload it to pub.dev because reflectable 4.0.11 is broken. If you have any concerns then I can create a new PR to address them.

sigurdm:
LGTM

eernstg:
Thanks!

PR #343, opened 2025-08-28
Update reflectable to work with Flutter 3.35

eernstg:
This PR changes the version of the analyzer to ^7.0.0, such that there is no dependency on `_macros` (which doesn't exist with Flutter 3.35, so we really need this step).

It makes minimal changes with respect to the new element model (`Element2`), just enough to make the code generator and generated code work in this new context. This implies that there will be another update where reflectable switches to the new element model (which will allow reflectable to use yet newer versions of the analyser).

This PR also prepares reflectable for being published as version 4.0.14.

### DIFF
--- a/reflectable/CHANGELOG.md
+++ b/reflectable/CHANGELOG.md
@@ -1,8 +1,11 @@
 ## 4.0.14
 
 - Use language version 3.7.
-
-
+- Migrate reflectable to use analyzer version ^7.0.0 (which removes
+  the dependency on `_macros`, which doesn't exist any more).
+- This includes a small part of the migration work needed in order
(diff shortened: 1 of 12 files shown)


eernstg:
@sigurdm, this PR was requested because Flutter 3.35 doesn't have `_macros`, and version 4.0.13 of reflectable dependencies are unable to avoid requesting `_macros`.

The switch to language version 3.7 caused a lot of reformatting, but `git diff -bw master...` can be used to avoid showing many of those changes. I can create an extra PR that only does reformatting and rebase this one on top of thet, if that's helpful.

sigurdm (approved):

PR #344, opened 2025-09-02
Use language and format version 3.9

eernstg:
This is a PR that only changes the language version and formatting of reflectable to 3.9. In particular, there are no changes to the semantics. It also prepares reflectable for release as version 4.0.15.


eernstg:
Thanks!

Issue #336, opened 2024-10-24
[Feature Request] separate Builder from types needed at runtime


devmil:
We currently face an issue where a library we use, uses reflectable and therefore transitively introduces the Builder bundled with reflectable.
We have a highly dynamic and huge package landscape and a couple of them use other Builders to generate code. 
The reflectable Builder is now causing issues (out of memory) and all those packages need to disable the reflector Builder in their build.yaml to continue working.

I think it would be better to have the Builder separated so that it can be used as a dev dependency not leaking it to all consumers.

d-silveira:
please, please, do like json_serializable and other packaged builders, and split reflectable into 2 different packages, 1 for the builder and another for the annotation, so we don't need to introduce a lot of unneeded dependencies beyond the dev dependencies declaration!

eernstg:
This one fell through the cracks, sorry about that. To get it restarted, it would be great if you could give me some further information.

> better to have the Builder separated

How would splitting reflectable into two packages make anything smaller? A library which is not part of the program of interest does not cost anything in terms of program size on disk or in memory, or in terms of compilation/analysis time, no matter which package that library resides in.

On the other hand, the fact that `reflectable.dart` imports `src/reflectable_builder_based.dart` cannot be changed by putting those two libraries in different packages: `src/reflectable_builder_based.dart` contains the code which is needed in order to use the code which is generated by the reflectable code generator. In other words, no program can use the generated code without containing `src/reflectable_builder_based.dart`.

Anyway, I hope you can clarify the situation a bit such that those out-of-memory incidents can be avoided!

eernstg:
Thanks for the input!

Would this be helpful?: https://github.com/google/reflectable.dart/pull/345. It's just a quick experiment, so there may be glitches.

devmil:
Looks good! I can try that tomorrow in our code base and give feedback

devmil:
@eernstg 
I was not able to directly test it in our codebase (too recent SDK version needed) but I re-created a comparable setup and it worked well 👍

eernstg:
Done!

PR #345, opened 2025-09-04
Split off the builder

eernstg:
This PR changes the `reflectable` package such that it only contains run-time related resources. The code generator is moved to a new package `reflectable_builder`.


eernstg:
@jakemac53, I can't add you as a reviewer, but it would be great if you could take a look at `reflectable_builder/build.yaml`. The change in this PR basically consists of moving files dealing with code generation from `reflectable` to `reflectable_builder`, such that clients can make `reflectable_builder` a `dev_dependency` and not a regular `dependency` (this has been requested in issue #336).

reflectable_builder/build.yaml
```diff
@@ -1,7 +1,7 @@
 builders:
   reflectable:
```
jakemac53, line 2: Consider renaming this to `reflectable_builder` as well, which will allow the shorthand syntax of just the package name to refer to it.

reflectable_builder/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
+name: reflectable_builder
+version: 5.0.0
+description: >
+  Reflection support based on code generation, using 'capabilities' to
+  specify which operations to support, on which objects. This is the
+  builder part of reflectable.
+homepage: https://github.com/google/reflectable.dart
+environment:
+  sdk: '>=3.9.0 <4.0.0'
```
jakemac53, line 9:
```suggestion
  sdk: ^3.9.0
```

reflectable_builder/pubspec.yaml
```diff
@@ -0,0 +1,28 @@
...
+dependencies:
+  analyzer: ^7.7.0
+  build: ^2.4.0
+  build_resolvers: ^2.4.0
+  build_config: ^1.1.0
+  build_runner: ^2.4.0
```
jakemac53, line 15: This should be a dev dependency, as well as many of the other ones probably

reflectable/pubspec.yaml
```diff
@@ -18,7 +19,7 @@ dependencies:
   logging: ^1.2.0
   package_config: ^2.1.0
   path: ^1.9.0
-  pub_semver: ^2.0.0
+  pub_semver: ^2.2.0
```
jakemac53, line 22: I can't comment on the correct line because GitHub is dumb but all the build related DEPS here should probably be dropped now - if this package does use build_runner the DEPS should be moved to dev dependencies at a minimum.

test_reflectable/pubspec.yaml
```diff
@@ -11,10 +11,14 @@ environment:
...
-    path: ../reflectable
+  reflectable: any
 dev_dependencies:
   analyzer: any
   build_test: any
   lints: any
+  reflectable_builder:
+    path: ../reflectable_builder
   test: ^1.25.0
+dependency_overrides:
```
sigurdm, line 22: Instead of this you could use a pub workspace to develop the two packages.

jakemac53:
The difference is transitive dependencies - builders typically come with a large amount of transitive deps, but they should really only be dev dependencies and not leak out to your users.

The annotations (and sometimes runtime libraries) are the only part that actually is imported under `lib/`, so putting those in a separate package allows you expose very minimal regular dependencies.

While it is true that probably tree shaking would reduce the total compiled _size_ of your applications, you still pay significant costs in terms of compilation time (simply importing package:build will result in a compile time of many seconds).

Then there is the increased risk of dependency conflicts (especially re: analyzer), and general bloat involved with having a large number of dependencies.

devmil:
I second @jakemac53 's comment.
Additionally to that: at least in our context builders that are referenced transitively are also executed by the build runner. This adds to the compilation time issue and is what was actually making us recognize this issue (as reflectable is a very expensive builder by nature and was bringing our dev machines and CI runners to their resource limits in our very big Flutter project)

sigurdm:
I'm not doubting that is how it works now, just wondering if that can be fixed.
But if we want to split the package anyway, it doesn't really matter.

eernstg:
The design of reflectable is inherently relying on being able to inspect a larger part of the program than the part which is reachable through imports from each library which is being processed.

For example, if package P1 uses P2, and P2 uses reflectable and has a reflector `reflector` then a class `C` in P1 may have the metadata `@reflector`, and reflectable must then generate reflection support code for `C`.

This means that if P2 relies on reflectable to support reflection on anything written by clients of P2 then it must have a broader perspective on the program than the usual analysis scope (that is, the current library plus everything which is directly or indirectly imported), which is basically the same thing as saying that reflectable is inherently based on a closed-world assumption. As I see it, that's the reason why it is natural for reflectable to be executed on the entire program, and hence it should run if reflectable is used by any dependency, direct or indirect.

The other side of the coin is that the closed-world assumption allows us to do things which are otherwise not supported: For example, we can iterate over all subtypes of a given class (using a `@subtypeQuantifyCapability`), even the ones that aren't reachable by any import path from the location where we're performing this iteration.

A package P4 could rely on a more narrow scoping where reflection is only supported on declarations in P4 and its dependencies, but if a package P3 imports something from P4 then it will be able to use reflection on P4 things (but not on anything in P3, not even on an instance of a P3 class that implements or extends a class in P4). This could work, but it requires caution, and it hasn't been explored very intensively (as far as I know).

So it isn't a bug to run code generation in the context of a client of a package that uses reflectable, it's by design.

sigurdm:
> So it isn't a bug to run code generation in the context of a client of a package that uses reflectable, it's by design.

I think there are (at least) two ways to use reflectable when you a making a package for others to depend on:

1. reflect on the package itself, and bake that into the package by committing the generated file(s), and importing them from the package. That gives less than a whole world view of the elements when the package is being consumed. I don't know how well this is supported currently. But that is where it makes sense to split the package into a generator and a builder
2. By asking the consuming package to run the generator over the whole world, and include the generated code in the main app. In this case you gain nothing by splitting the package, as the app author would always end up having to depend on the builder package.

test_reflectable/pubspec.yaml
eernstg: OK, using a pub workspace now, PTAL.

eernstg:
Thanks for the responses!

gmpassos:
> @gmpassos, could you give some more details about how this change would help you?

**Real-world case:**

* We have a project with two packages: `api` (backend) and `ui` (Web UI).

  * The `api` package uses `reflection_factory` (a reflection package similar to `reflectable`, which generates reflection code).
  * The `ui` package depends on `api` and also pulls in `reflection_factory` and `build` as transitive dependencies.

    * To avoid unnecessary build steps, we configure `build.yaml` with `reflection_factory/enabled: false`.
    * Over the past year, we’ve had some issues with transitive dependencies—especially with `analyzer`, since `reflection_factory` lagged behind the latest `analyzer` release for about a month.

IMHO, since the reflection package is only used in the `api` package for code generation, the `ui` package shouldn’t be affected or forced to load unnecessary transitive dependencies. A major challenge in medium- to large-scale real-world projects is dependency management, so any approach that helps reduce this complexity is always welcome.

PR #348, opened 2025-09-30
Small fixes

eernstg:
Fix a link in `reflectable_builder/README.md`. Add an example to `reflectable` that doesn't depend on code generation (just shows the basic structure of the code). Format all tests (so it should not be necessary to look at the tests in detail).

Prepares `reflectable` for publishing v5.0.1 and `reflectable_builder` for v1.0.1.


packages/reflectable/example/example.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2025, the Dart Team. All rights reserved. Use of this
```
sigurdm, line 1: Alternatively you could make the example depend on the builder, and have its own pubspec.

I don't think this example is really helpful.

sigurdm: Yes - You can include it in the workspace.

sigurdm:
Yes: These are the ones we support: https://github.com/dart-lang/pana/blob/8451cc5d036c3338ee58139f80d0011ae0fc0d5e/lib/src/maintenance.dart#L9

PR #355, opened 2026-02-04
fix for issue 354

StuartSharpless:
Fix, if you want it, for issue 354 Exception in _getLibraryMetadata while running reflectable_builder

Just adding the length check for the directives list.

### DIFF
--- a/packages/reflectable_builder/lib/src/builder_implementation.dart
+++ b/packages/reflectable_builder/lib/src/builder_implementation.dart
@@ -5593,7 +5593,7 @@ CompilationUnit? _definingCompilationUnit(
 
 // Helper for _extractMetadataCode.
 NodeList<Annotation>? _getLibraryMetadata(CompilationUnit? unit) {
-  if (unit != null) {
+  if (unit != null && unit.directives.isNotEmpty) {
     Directive directive = unit.directives[0];
     if (directive is LibraryDirective) {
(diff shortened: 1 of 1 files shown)


eernstg (approved):
Thanks @StuartSharpless, this looks good to me!

@sigurdm, could you take a look as well?

eernstg:
@StuartSharpless, the 'contributor license agreement' needs to be in place in order to land this PR. If you're OK with that then please check out the 'Details' in the 'cla/google' item above. Otherwise I can just create a new PR to do the same thing.

StuartSharpless:
@eernstg I think I have now completed the CIA it shows successful for me. I thought I had before as well :-)

eernstg:
It's such a basic thing to check that a list isn't empty before accessing the first element in it, so it's kind of surprising that this bug could exist for years and years without being detected.

However, it does make sense that it wouldn't happen often: The empty list can only be encountered in a situation where the reflectable code generator is invoked on an entry point that has no imports (and no exports, etc, but "no imports" is the one that matters), and this implies that the run would fail very soon after: The code generator bails out if the program (all libraries reachable from the entry point) doesn't include `reflectable.dart`. So we'll avoid the crash, but the code generator will then immediately end the run and announce that there's nothing it can do with this entry point. ;-)

sigurdm (approved):

PR #356, opened 2026-02-05
Prevent generated imports of `dart:_http`

### LINKED ISSUE #325: importing private libaries during build_runner build
```dart
// This file has been generated by the reflectable package.
// https://github.com/dart-lang/reflectable.

import 'dart:core';
import 'dart:_http' as prefix8; // <---- why is it importing private libraries? 
import 'dart:async' as prefix9;
import 'package:pharaoh/src/http/request.dart' as prefix5;
import 'package:pharaoh/src/http/response.dart' as prefix6;
```

I am trying to use Cookie in a Class that is reflectable. I imported Cookie from 
```dart
import 'dart:io' show Cookie;
```

And this is the error i get 
```shell
test/router_test.reflectable.dart:5:8: Error: Can't access platform private library.
  import 'dart:_http' as prefix9;
```

eernstg:
This PR improves on the situation which is reported in #325 (which is that reflectable can generate code that doesn't compile by importing an internal platform library) by special casing `dart:_http`. It is not obvious that there is a general solution, but `dart:_http` is _only_ exported by `dart:io` which implies that we can always replace any import of `dart:_http` by an import of `dart:io`.

Also correct some URLs in the top-level README.md.

Also prepare reflectable/builder to be released as version 5.2.1/1.2.1.

Fixes #325


eernstg:
I've asked for an example in #325 which would allow the new code to be tested.

packages/reflectable_builder/lib/src/builder_implementation.dart
```diff
@@ -4897,6 +4897,27 @@ class BuilderImplementation {
...
 
+    // Some "private" platform libraries are exported by public ones. To avoid
+    // the error caused by importing 'dart:_http', we transform it to 'dart:io'.
+    // Other cases are ambiguous and not handled.
+    final length = imports.length;
+    for (int index = 0; index < length; ++index) {
+      final import = imports[index];
+      if (import.contains('dart:_')) {
+        if (import.contains('dart:_http')) {
+          final rest = import.substring(20, import.length);
```
sigurdm: The 20 looks like a magic constant.

Could you do (if that is what is happening):

```suggestion
          final rest = import.substring("import 'dart:_http' ".length, import.length);
```

packages/reflectable_builder/lib/src/builder_implementation.dart
```diff
@@ -4897,6 +4897,27 @@ class BuilderImplementation {
...
     imports.sort();
 
+    // Some "private" platform libraries are exported by public ones. To avoid
+    // the error caused by importing 'dart:_http', we transform it to 'dart:io'.
+    // Other cases are ambiguous and not handled.
+    final length = imports.length;
+    for (int index = 0; index < length; ++index) {
+      final import = imports[index];
+      if (import.contains('dart:_')) {
+        if (import.contains('dart:_http')) {
```
sigurdm: Perhaps be more specific, to avoid a theoretical case of a filename containing "dart:_"
```suggestion
      if (import.contains("'dart:_")) {
        if (import.contains("'dart:_http'")) {
```

packages/reflectable_builder/lib/src/builder_implementation.dart
eernstg: Makes sense, done!

packages/reflectable_builder/lib/src/builder_implementation.dart
eernstg: Indeed, done!

sigurdm (approved):
Ideally we find the "canonical" library that exports a given element instead of special casing on imports after they have been generated.

Something like what we do in dartdoc: https://github.com/dart-lang/dartdoc/blob/cb8f442d36abb611d67c8d2fb3e702a641ea32dd/lib/src/model/model_element.dart#L503

Maybe it is ok to land this as a temporary fix though.

eernstg:
OK! I actually changed the code generator in a slightly different way: It consistently treats 'dart:_http' as an importable library all the way to the end (which is needed in order to ensure that we will generate code for a `Cookie` class mirror and such). At the very end, the import of 'dart:_http' is changed (as before) to import 'dart:io' instead.

It may seem hacky to do this at the textual level, but I actually think it's a consistent approach.

If I change the construction of domains such that declarations in 'dart:_http' pretend to be declared in 'dart:io' then various other things are breaking.

PTAL!

sigurdm (approved):

eernstg:
Thanks!

PR #358, opened 2026-05-19
Update versions to language 3.12 and analyzer 11

eernstg:
This PR updates the dependencies in all packages in this repository to use language version 3.12, and it updates the dependencies on the analyzer (where present) to ^11.0.0. It prepares reflectable for being released as version 5.1.2 and reflectable_builder as version 1.2.2.

### DIFF
--- a/packages/reflectable/CHANGELOG.md
+++ b/packages/reflectable/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.2.2
+
+- Update the language version to 3.12.
+
 ## 5.2.1
 
 - Fix bugs #354 and #325.
--- a/packages/reflectable/pubspec.yaml
+++ b/packages/reflectable/pubspec.yaml
@@ -1,12 +1,12 @@
 name: reflectable
-version: 5.2.1
+version: 5.2.2
 description: >
   Reflection support based on code generation, using 'capabilities' to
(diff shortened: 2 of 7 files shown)


eernstg:
Hi @sigurdm, reflectable was at analyzer version ^10.0.0 and the analyzer is now at ^13.0.0, so I'm migrating the code to get aligned with the new analyzers.

Going from ^10.0.0 to ^11.0.0 is mainly about the breaking change where the property `isSynthetic` is eliminated, and various subtypes of `Element` instead gained one or more properties which are more specific. For example, a `PropertyAccessorElement` has an `isOriginVariable` which is true if and only if this property accessor declaration was created implicitly based on a variable declaration.

Note that the property `isOriginDeclaration` is arguably interpreted in the opposite direction: This one is true if _this_ declaration is "the original one" (that is, it wasn't created implicitly based on some other declaration).

This means that it is necessary to be careful about which new property to use in place of the old `isSynthetic` which isn't available any more. In general `isSynthetic` is replaced by `isOriginVariable` or `!isOriginDeclaration` (note the negation), but in some cases there is no such property and the status is hardcoded (e.g., a class element doesn't have those properties, presumably because it is never synthetic).

sigurdm (approved):

eernstg:
Thanks!

PR #359, opened 2026-05-20
Update dependency to use analyzer 13

eernstg:
This PR changes the dependencies in the reflectable packages to use analyzer ^13.0.0 and migrates the code generator accordingly. In particular:

Analyzer ^13.0.0 removes the class `DefaultFormalParameter`. Default values are now handled via a property on `FormalParameter`. Old approach: Check if a parameter node is a `DefaultFormalParameter`, access its `defaultValue`. New approach: Check the `defaultClause` property on a `FormalParameter` or `RegularFormalParameter` and access its `value`.

Similarly, the class `NamedExpression` has been removed and replaced by `NamedArgument`, in order to model named arguments in an `ArgumentList`. Old approach: Use `expression is NamedExpression` and access `expression.name` and `expression.expression`. New approach: Use `expression is NamedArgument` and access `expression.name.lexeme` to get the label name and `expression.argumentExpression` to get the value.

Finally, the `arguments` property of `ArgumentList` now returns a `NodeList<Argument>` instead of a `NodeList<Expression>`. Various loops have been updated to use the new types.

This PR also prepares `reflectable` for a release as v5.2.3 and `reflectable_builder` as v1.2.3. Version v5.2.3 of `reflectable` doesn't actually change anything other than the version, but I'd like to allow all usages of `reflectable` and `reflectable_builder` to remain in sync with respect to the version number such that developers can always update those two IDs together (e.g., from 5.2.2 to 5.2.3 along with 1.2.2 to 1.2.3, simultaneously).

Finally, this PR fixes a bug in 'example.dart' (where a lint asked for a type annotation on a parameter, but the type annotation which was added was wrong).

### DIFF
--- a/packages/reflectable_builder/lib/src/builder_implementation.dart
+++ b/packages/reflectable_builder/lib/src/builder_implementation.dart
@@ -2731,15 +2731,15 @@ class _ReflectorDomain {
             as FormalParameter?;
     // The node can be null because the declaration is synthetic, e.g.,
     // the parameter of an induced setter; they have no default value.
-    if (parameterNode is DefaultFormalParameter &&
-        parameterNode.defaultValue != null) {
+    FormalParameterDefaultClause? defaultClause = parameterNode?.defaultClause;
+    if (defaultClause != null) {
       return await _extractConstantCode(
-        parameterNode.defaultValue!,
+        defaultClause.value,
         importCollector,
         _generatedLibraryId,
         _resolver,
       );
-    } else if (parameterElement is DefaultFormalParameter) {
+    } else {
       Expression? defaultValue = parameterElement.constantInitializer;
       if (defaultValue != null) {
         return await _extractConstantCode(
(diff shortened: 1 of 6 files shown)


eernstg:
Hello @sigurdm, the update to analyzer 12.0.0 turned out to be non-breaking (as seen from this package, at least), so I took another step and updated it directly to analyzer 13.0.0.

packages/reflectable_builder/lib/src/builder_implementation.dart
```diff
@@ -5531,19 +5532,19 @@ Future<String> _extractConstantCode(
...
+    } else if (expression is NamedArgument) {
       String value = await _extractConstantCode(
-        expression.expression,
+        expression.argumentExpression,
         importCollector,
         generatedLibraryId,
         resolver,
       );
-      return '${expression.name} $value';
+      return '${expression.name.lexeme}: $value';
```
sigurdm, line 5547: Not sure what is happening here, did we get an extra colon?

sigurdm (approved):

packages/reflectable_builder/lib/src/builder_implementation.dart
eernstg: `expression.name` used to be a `Label` whose `toString()` includes the colon, but the new type of `expression.name` is `Token` and there's a separate instance variable in `expression` holding the colon token. I don't think there's a more concise way to get the right data here, because the `value` needs to be obtained from `_extractConstantCode` (so we can't just print the whole `NamedArgument`.

PR #360, opened 2026-05-28
Add dependabot configuration to packages reflectable and reflectable_builder

eernstg:
This PR adds a dependabot configuration file to this repository, requesting version updating PRs for the packages such that they can more easily be kept up to date with respect to dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,6 @@
 version: 2
 updates:
-  # Configuration for packages/reflectable
   - package-ecosystem: "pub"
-    directory: "/packages/reflectable"
+    directory: "/"
     schedule:
       interval: "weekly"
-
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+updates:
+  # Configuration for packages/reflectable
+  - package-ecosystem: "pub"
+    directory: "/packages/reflectable"
```
sigurdm: Because everything is a single workspace, I think you just need a single directory entry pointing at the root.

See https://github.com/dart-lang/pub/pull/4294

eernstg: OK, thanks! I simplified the configuration, PTAL.

sigurdm (approved):

